PR #35, opened 2023-10-18
feat: write output files to directory

### LINKED ISSUE #13: Write output to files
Instead of defaulting to the screen, let's write the output to files in a directory

wolf31o2:
Closes #13

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -18,6 +18,9 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"path/filepath"
+
+	"golang.org/x/sync/errgroup"
 
 	"github.com/blinklabs-io/bursa"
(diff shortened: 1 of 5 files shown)


internal/cli/cli.go
```diff
@@ -58,5 +61,60 @@ func Run() {
...
+						file, err := os.Create(path)
+						if err == nil {
+							defer file.Close()
+							_, err := file.Write([]byte(v))
+							if err != nil {
+								return err
+							}
+						}
+						return err
+					}
```
agaffney, line 100: Creating the output directory inside multiple concurrent goroutines is a bad idea. This should be done once, outside of this loop

wolf31o2: It's not done inside a loop.

agaffney (approved):

Issue #63, opened 2024-02-04
Use cobra for commandline handling


agaffney:
It handles subcommands and supports shell completion

PR #89, opened 2024-05-21
feat: add cobra cli library

oversize:
This PR adds cobra as the cli parsing library. https://github.com/blinklabs-io/bursa/issues/63

- I have copied the way cardano-up has used cobra more or less exactly
- The cli currently has the same ability as before. I wanted to start some better cli then "cli" but felt that it would be better to have some more discussion before implementing it. I will add my thoughts/ideas there 
- I removed internal/logging in favour of internal/consolelog also stolen from cardano-up
- I needed to remove some loggig output from the api, because i dont know gin and did not spend time figuring out how to adapt slog to work with it. Also: I would like to suggest to remove gin and use  go's 1.22 new stdlib net/http muxer instead. Removing the dependency alltogether. The api seems fairly simple so i think net/http is all we need. But i will open. another ticket for that discusson.
-


wolf31o2:
Looks like you have conflicts with main.

oversize:
> Looks like you have conflicts with main.
Oh snap, yes. I have merged these changes now and updated the commits.
I also just saw that cobra still provided its default "generate shell completion" script, which i now have disabled.

cmd/bursa/new.go
```diff
@@ -16,8 +16,34 @@ package main
...
 
-func cliMain() {
-	cli.Run()
+var (
+	output string
+)
+
+func newCommand() *cobra.Command {
+	newCommand := &cobra.Command{
+		Use: "new",
```
wolf31o2, line 28: `bursa new wallet` is describing a state vs an action. `bursa create wallet` is giving and action, but I think I prefer `bursa wallet create` where the immediate first sub-command is the subject rather than the action. It'll also more closely reflect the API path `/api/wallet/create`

internal/api/api.go
```diff
@@ -17,17 +17,16 @@ package api
 import (
 	"fmt"
 	"net/http"
-	"time"
 
-	ginzap "github.com/gin-contrib/zap"
+	"log/slog"
```
wolf31o2, line 21: You'll need to update the `.github/workflows/go-test.yml` and `.github/workflows/publish.yml` to use 1.21/1.22 (build using 1.21 for now) if you want to update this. Personally, I would split the logging changes from the CLI changes and do this in another pull request.

internal/api/api.go
```diff
@@ -58,16 +57,14 @@ func Start(cfg *config.Config) error {
...
-		SkipPaths:  []string{},
-	}))
-	router.Use(ginzap.RecoveryWithZap(accessLogger, true))
+	// accessLogger := logging.GetAccessLogger()
+	//router.Use(ginzap.GinzapWithConfig(accessLogger, &ginzap.Config{
+	//	TimeFormat: time.RFC3339,
+	//	UTC:        true,
+	//	SkipPaths:  []string{},
+	//}))
+	//router.Use(ginzap.RecoveryWithZap(accessLogger, true))
```
wolf31o2, line 67: Access logs in the API are a requirement for us. We cannot just remove them.

wolf31o2 (changes requested):
Back out the changes to remove the `internal/logging` module. I'm fine with the introduction of slog for console logging, but we need the access logs in the API to stay.

oversize:
> Back out the changes to remove the `internal/logging` module. I'm fine with the introduction of slog for console logging, but we need the access logs in the API to stay.

I have removed the console logging again in favor of internal/logging. I'll look into gin and how to have access logs with log/slog so we dont lose these when switching to it.

cmd/bursa/new.go
oversize: Just pushed a change that does that; Rename `new` command to wallet and add a create subcommand in that so that its `bursa wallet create` now.

wolf31o2 (approved):

PR #554, opened 2026-06-29
feat(connector): CIP-30 + CIP-95 dApp connector (backend + SPA + MV3 extension)

wolf31o2:
pr554-connector

---
## Summary by cubic
Adds an opt-in CIP‑30 + CIP‑95 dApp connector across backend, SPA, and an MV3 extension. The extension injects `window.cardano.bursa`, relays to the local Bursa instance, and enforces per‑site consent with in‑app approvals.

- **New Features**
  - Backend: `ui/internal/connector` with pairing token store, per‑origin grants, approval queue (SSE), and service/dispatcher; `/connector/*` is guarded by token + strict CORS and only enabled with `BURSA_CONNECTOR=1` (`api.WithConnector` binds it to the active wallet).
  - CIP‑30/95: read methods (network id, UTxOs/balance/addresses/collateral), `signTx`, `signData`, `submitTx`; CIP‑95 `getPubDRepKey`, stake key queries; spend derives the DRep key and includes stake/DRep keys as required signers.
  - SPA: global approval overlay via SSE with Web Notifications; Settings shows pairing state, pending pairings, connected sites (revoke), and unpair; dev proxy for `/connector`.
  - Extension (MV3): injected/content/background provider exposing `window.cardano.bursa`, popup pairing UI, exact‑origin relay (origin from `sender.origin`), manual E2E sample dApp; background requests use a 125s timeout; CI lints/tests/builds `ui/extension` and uploads `bursa-connector-dist`.

- **Bug Fixes**
  - Accept `cip95.*` method names in the Go dispatcher; align `isEnabled()` error shape with CIP‑30.
  - Harden origin/CORS checks against DNS rebinding; only `/connector/pair` is exempt and must match exactly. Content replies only to the page’s exact origin, handles opaque origins, and returns structured errors if the MV3 worker is unavailable.
  - Approval queue uses timeouts and caps pending requests per origin; SPA hooks add `enabled` to avoid unnecessary polling in idle views.
  - Vault: add `VerifyPassword` so approvals can authenticate without unlocking.

<sup>Written for commit 96505e22b130976054411b188b804044a9a3a5c1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/bursa/pull/554?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added browser extension support for pairing, approving, and managing dApp connector requests.
  * Added a new in-app connector approval overlay, plus settings controls to view paired status, pending pairings, connected sites, revoke access, and unpair.
  * Added connector request notifications and a direct test harness for extension verification.

* **Bug Fixes**
  * Improved request handling, origin validation, and timeout behavior for connector interactions.
  * Strengthened extension and app build/test coverage for more reliable releases.


ui/internal/api/connector.go
```diff
@@ -0,0 +1,580 @@
...
+}
+
+// handleConnectorGrants handles GET /connector/grants.
+//
+// SPA-facing — requires same-origin request (no Origin header or Origin matches
+// the server's own scheme+host). Returns the list of granted origins plus the
+// paired extension ID if one exists.
+func handleConnectorGrants(svc *connector.Service) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if !sameOrigin(r) {
```
chrisguiney, line 196: **major (proven bypass):** this handler uses the lax `sameOrigin`, while every other SPA-facing route (events, grants/revoke, decide, unpair, pending-pairings) uses `strictSameOrigin`. Combined with `sameOrigin` returning true on an absent Origin header (see line 120), this is DNS-rebindable.

Browsers omit `Origin` on same-origin GETs, so a page served from `evil.com` that rebinds `evil.com → 127.0.0.1` can do `fetch("/connector/grants")` and read back the paired extension ID, paired state, and the full list of connected dApp origins. I proved it with a scratch test: a request with `Host: evil.com:8090` and no Origin passes `sameOrigin` (returns true), while `strictSameOrigin` correctly rejects the identical request.

Read-only metadata — no keys, no signing — but it defeats the DNS-rebind defense the comments on `sameOrigin` explicitly claim. **Fix:** switch this guard to `strictSameOrigin` (matching the sibling routes), or make `sameOrigin` enforce `isLoopbackHost(r.Host)` even when Origin is absent.

ui/internal/api/connector.go
```diff
@@ -0,0 +1,580 @@
...
+
+// sameOrigin returns true when the request comes from the same origin as the
+// API server. Same-origin browser requests typically omit the Origin header
+// entirely, so an absent header is treated as same-origin. When Origin is
+// present, the Host must be a loopback address (DNS-rebinding defence) and the
+// Origin must match the API server's own scheme+host.
+func sameOrigin(r *http.Request) bool {
+	o := r.Header.Get("Origin")
+	if o == "" {
+		// No Origin header: same-origin direct navigation, allow.
```
chrisguiney: **root cause of Finding 1:** returning `true` here, before the loopback-Host check below, is what makes the absent-Origin case rebindable. The comment ("same-origin direct navigation, allow") is true for the SPA's own no-Origin GETs, but a DNS-rebound cross-site page produces exactly the same shape (no Origin, relative fetch). `strictSameOrigin` exists precisely to close this and is used everywhere except `handleConnectorGrants`. Enforcing `isLoopbackHost(r.Host)` on this branch too would harden `sameOrigin` itself rather than relying on each caller to pick the strict variant.

ui/web/src/screens/ConnectorApproval.tsx
```diff
@@ -0,0 +1,309 @@
...
+  useEffect(
+    () => () => {
+      previouslyFocused.current?.focus();
+    },
+    [],
+  );
+
+  if (!current) return null;
+
+  async function handleDecide(approved: boolean) {
```
chrisguiney, line 207: **major (consent-integrity):** `handleDecide` is bound to `current = pending[0]` (line 123), recreated each render. If the head is removed externally — a server-side timeout, or a decision made in another tab — just as the user clicks, the next request instantly takes the same button position and the click lands on a request the user never reviewed.

I verified the mitigation boundary: the `[currentID]` effect clears the password and Approve is `disabled={busy || (requiresPassword && !password)}`, so **all key-signing paths are protected** (they can't be one-click approved). The residual is the two no-password methods — `enable` (grants a persistent origin connection) and `submitTx` (broadcasts an already-signed tx). So this is narrower than "approve anything," but `enable` being silently substitutable is still a real gap.

**Fix:** on `currentID` change, gate Approve behind a fresh interaction for all methods (extend the password-clear pattern), or show a "request changed — re-review" state so a swap can't convert a click into an approval.

ui/web/src/screens/ConnectorApproval.tsx
```diff
@@ -0,0 +1,309 @@
...
+                {current.origin}
+              </code>
+            </dd>
+            <dt>Action</dt>
+            <dd>{methodSummary(current)}</dd>
+            {current.params !== undefined && (
+              <>
+                <dt>Params</dt>
+                <dd>
+                  <pre
```
chrisguiney, line 293: **major (informed consent):** the overlay does display the actual payload from the authoritative SSE snapshot (good), but `formatParams` renders it as raw hex/JSON. For `signTx`/`signData` a user cannot tell whether the tx sends 5 ADA or drains the wallet — the approval gates consent but not *informed* consent, which is the whole point of the step. I confirmed there's no backend decode/preview path anywhere in the connector.

This likely needs a backend decode endpoint (the SPA has no ledger to interpret CBOR). At minimum, render a prominent "contents unverified — raw transaction" warning next to the payload so the raw hex isn't mistaken for a reviewed summary. (Rendering itself is safe — this is a `<pre>` text child, so no injection concern.)

ui/internal/api/connector_test.go
```diff
@@ -0,0 +1,1371 @@
...
+			req.Host = "localhost:8080"
+			rec := httptest.NewRecorder()
+			mux.ServeHTTP(rec, req)
+			if rec.Code != http.StatusForbidden {
+				t.Errorf("%s %s with evil Origin: status = %d, want 403", p.method, p.path, rec.Code)
+			}
+		}
+	})
+
+	t.Run("same-origin guard: GET grants no Origin header → allowed", func(t *testing.T) {
```
chrisguiney: **major (test blesses the bug):** this case asserts that a no-Origin GET to `/connector/grants` returns allowed — codifying Finding 1's leak as intended behavior. The DNS-rebind test elsewhere only exercises the Origin-*present* case, so the leak sits inside the tested-and-blessed surface rather than an untested corner. When you switch grants to `strictSameOrigin`, this expectation should flip to "rejected," and it's worth adding an explicit rebind case: `Host: evil.com`, no Origin → 403.

ui/internal/connector/service.go
```diff
@@ -0,0 +1,501 @@
...
+// bearer token. Returns ErrPairCodeMismatch if the code is wrong.
+func (s *Service) ConfirmPair(extensionID, code string) (string, error) {
+	extensionID = normalizeExtensionID(extensionID)
+	s.pairMu.Lock()
+	expected, ok := s.pairCodes[extensionID]
+	if ok && expected == code {
+		delete(s.pairCodes, extensionID)
+	}
+	s.pairMu.Unlock()
+	if !ok || expected != code {
```
chrisguiney: **minor:** the pairing code — the bootstrap secret that gates token minting — is compared with plain `==`/`!=` (line 135 and here), which short-circuits at the first differing byte (a timing side channel), and `pairCodes` entries have no expiry or attempt cap. The token itself is correctly `subtle.ConstantTimeCompare`d, so this is the weaker link. Keyspace (10^12), loopback timing noise, and the fact that a minted token still can't sign without the per-request SPA password keep it minor — but a local process could timing-attack or brute-force the code. **Fix:** `subtle.ConstantTimeCompare` the code, add an expiry, and cap attempts.

ui/internal/connector/service.go
```diff
@@ -0,0 +1,501 @@
...
+// account clears the binding.
+func (s *Service) SetActiveAccount(walletID string, acct *wallet.Account) {
+	if rb, ok := s.be.(accountRebinder); ok {
+		rb.SetAccount(walletID, acct)
+	}
+}
+
+// BeginPair generates a numeric pairing code for the given extensionID and
+// caches it. The code is returned for display in the Bursa UI.
+func (s *Service) BeginPair(extensionID string) string {
```
chrisguiney: **minor:** `BeginPair` stores one entry per `extensionID` with no format validation, no cap, and no expiry, and `/connector/pair` is token-exempt. A local process (or possibly a web page via a no-preflight simple POST, depending on Chrome Private Network Access enforcement) can add unlimited distinct entries — memory growth plus flooding the Settings "pending pairings" list. The codes stay password-gated so it's DoS/social-engineering noise, not an auth bypass. **Fix:** bound and expire the map, and validate the extension ID here.

ui/internal/connector/queue.go
```diff
@@ -0,0 +1,194 @@
...
+	defer q.remove(id)
+	// Use NewTimer + defer Stop to avoid a goroutine/channel leak when the
+	// request is resolved before the timeout fires (time.After leaks the
+	// timer until it fires at the full 120s production timeout).
+	t := time.NewTimer(q.timeout)
+	defer t.Stop()
+	select {
+	case d := <-w.done:
+		return d, nil
+	case <-t.C:
```
chrisguiney, line 92: **minor (fails safe):** if `Await` selects `<-t.C` at the same instant `Decide` sets `decided` and sends on the buffered `done` channel, `Decide` returns nil (SPA sees "approved") while `Await` returns `ErrTimeout` and the signing `fn` is never called. It fails safe — no signing occurs on a timeout — but leaves the SPA/dApp showing "approved" for a request that actually timed out. Worth a post-timeout re-check of `decided`, or at least a comment noting the intended resolution.

ui/extension/src/injected.ts
```diff
@@ -0,0 +1,177 @@
...
+      }
+    };
+
+    const timer = setTimeout(() => {
+      cleanup();
+      reject({ code: -2, info: 'No response from Bursa extension' });
+    }, REPLY_TIMEOUT_MS);
+
+    window.addEventListener('message', handler);
+    window.postMessage({ source: 'bursa-cip30', id, method, params }, '*');
```
chrisguiney: **nit:** requests are posted with `targetOrigin: '*'`. Posting to the current window only reaches same-window listeners (never child iframes or the opener), so this isn't a cross-origin leak — but `window.location.origin` here would match the rigor of the reply path and is the best-practice default. Separately, the redundant page-controlled `origin` carried in `params` (for `enable`/`isEnabled`) is inert today since the backend authorizes on `body.Origin`, but it's a latent footgun worth dropping before any future params struct trusts it. And the in-code comment implying the per-call nonce defends against same-origin co-resident scripts should be softened — it defends against other-window forgers only; a same-origin script (or XSS) can always observe and race the reply, which is the inherent floor of the injected-provider pattern, not a defect here.

chrisguiney (changes requested):
Reviewed across the four trust boundaries — Go connector backend, signing/vault, MV3 extension, and the React approval UI — verifying the load-bearing claims directly, including proving the one real bypass with a scratch test.

**This is a careful, security-aware PR, and the parts that would be catastrophic to get wrong are correct.** Stating the important positives up front because they check out end-to-end:

- **signData cannot be replayed as a tx witness.** `SignData` produces a COSE_Sign1 `Sig_structure` (`Ed25519(CBOR["Signature1", protected, [], payload])`) with the address bound into the protected header; a Cardano witness signs exactly the 32-byte `Blake2b256(txbody)`. The signData preimage is a CBOR 4-element array embedding the `"Signature1"` context, always much longer than 32 bytes, so the domains are provably disjoint — even a `payload` equal to a real body hash stays wrapped and unusable as a witness. This is the canonical CIP-8 footgun and it's handled correctly.
- **signTx re-derives the hash from the body it submits** (never trusts a client hash/witness); `partialSign` relaxes only completeness checks and can't expand the signed set; fails closed with `ErrWalletChanged` on a mid-approval wallet switch.
- **CIP-95 returns only 32-byte public keys** — chain code sliced off, correct soft-derived role-3/role-2 paths, tested against independently derived vectors.
- **`VerifyPassword`** zeroes the VEK on every path (nil-safe), fails via AES-GCM tag rather than an early `==`, holds only RLock, changes no lock/active-wallet state.
- **Origin trust chain is airtight**: the backend authorizes on `body.Origin`, set by the privileged extension worker from the browser's `sender.origin`; the page-controlled `params.origin` is never consulted, and a web page can't reach `/connector/request` (needs token + CORS). Token is 32-byte crypto/rand, constant-time compared, never in the page world, no SSRF path. CORS is non-reflective with no `Allow-Credentials`. Queue caps are enforced and a slow SSE consumer can't wedge signing.

All Go tests pass including `-race`; the web suite is 346/346 with a clean build; CI wires up lint/type-check/test/build for both `ui/web` and `ui/extension`.

Requesting changes for the items inline. The one I'd fix first — **Finding 1, a DNS-rebinding metadata leak on `GET /connector/grants`** — is a one-line fix (`strictSameOrigin`) and I proved it with a scratch test. The two approval-UI consent gaps (request-substitution race, raw-CBOR display) are the others I'd treat as blockers for a signature-gating flow, though the key-signing paths are already protected by the password-clear.

Two notes that don't fit inline:

- **Pre-existing, adjacent to this PR (not in the diff):** `spend.go` `Confirm` (~line 859) is the sole signing path that doesn't zero `rootKey`/`acctKey` — every sibling method defers zeroing the account master key. Best-effort severity, but worth fixing for consistency while this code is warm. Also pervasive: the mnemonic is copied into un-zeroable Go strings via `GetRootKeyFromMnemonic(string(mnemonicBytes), …)` on every signing path.
- **Worth escalating separately:** the non-connector Bursa API (`/vault/status`, `/wallet/*`) has no same-origin guard at all, a larger DNS-rebind exposure than Finding 1 that this PR's defense only partially addresses.

(`DATABASE.md`/`ARCHITECTURE.md` are dingo-repo conventions and don't apply to the bursa checkout.)

chrisguiney (approved):

PR #561, opened 2026-06-30
feat(ui): mobile-resilience wave — network/wake robustness + foreground service

wolf31o2:
---
## Summary by cubic
Adds Android and iOS shells and hardens network/wake behavior so the wallet stays in sync in the background and recovers cleanly after network changes or app resume. Unifies boot across desktop/mobile and ships an offline-aware web UI with safe retries and an offline banner.

- **New Features**
  - `ui/mobile`: `gomobile` binding (`App.Start/Stop/Port/OnNetworkChanged/OnResume`) that boots the wallet and serves the SPA on `127.0.0.1`.
  - Android/iOS shells: WebView apps loading the embedded SPA. Android runs a foreground `WalletService` (`dataSync`), shows a notification, listens for connectivity changes, and allows loopback cleartext via a strict `network_security_config`. iOS uses `WKWebView`, monitors reachability (`NWPathMonitor`), and links a generated xcframework via `xcodegen`.
  - Network/wake resilience: `supervisor.Reconnect` stop-then-relaunch, `boot.App` hooks for resume/network-change, plus unit tests around cancellation and relaunch.
  - Web UI robustness: read-only request retries (GET/HEAD/OPTIONS only) on network failures with capped backoff; polling pauses when offline/hidden and immediately refetches on online/visibilitychange/resume; adds an offline banner.
  - Unified boot: `ui/internal/boot` centralizes node/services/settings and binds an OS-assigned loopback port; desktop now calls `boot.Boot`. CI adds a `mobile` workflow that builds Android (canonical Docker image/script) and iOS (macOS runner) on tags or manual runs.

- **Migration**
  - Android: declare `FOREGROUND_SERVICE` with `foregroundServiceType="dataSync"` and add `FOREGROUND_SERVICE_DATA_SYNC` on API 34+.
  - Builds: add `golang.org/x/mobile` and the tool `golang.org/x/mobile/cmd/gobind`; Android bind targets `android/arm64`. Use the provided Docker image/script for Android or the CI workflow.

<sup>Written for commit dc397a75744dc9a373ddaaa4ddaa31560787647c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/bursa/pull/561?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added mobile app build and CI support for Android and iOS, including packaged artifacts for both platforms.
  * Introduced mobile app shells that keep the wallet available in the background and recover after network/app resume events.
  * Added offline-aware UI behavior, including an offline banner and immediate refetch on reconnect.

* **Bug Fixes**
  * Improved request handling with retries for transient network issues.
  * Added safer startup, shutdown, and resume behavior across mobile and web flows.

* **Documentation**
  * Added mobile build and architecture guidance for contributors.

### DIFF
--- a/mobile/android/app/src/main/java/io/blinklabs/bursa/WalletService.kt
+++ b/mobile/android/app/src/main/java/io/blinklabs/bursa/WalletService.kt
@@ -1,0 +1,358 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package io.blinklabs.bursa
+
+import android.app.Notification
+import android.app.NotificationChannel
+import android.app.NotificationManager
+import android.app.Service
+import android.content.Context
+import android.content.Intent
+import android.content.pm.ServiceInfo
+import android.net.ConnectivityManager
+import android.net.Network
+import android.net.NetworkCapabilities
+import android.net.NetworkRequest
+import android.os.Binder
+import android.os.Build
+import android.os.Handler
+import android.os.IBinder
+import android.os.Looper
+import androidx.core.app.NotificationCompat
+
+import io.blinklabs.bursa.mobile.App
+import io.blinklabs.bursa.mobile.Mobile
+import java.util.concurrent.Executors
+import java.util.concurrent.RejectedExecutionException
+
+// WalletService is a bound foreground Service that OWNS the gomobile App (the
+// embedded lean Dingo node + loopback control surface). Hosting the wallet in a
+// foreground service — rather than in the Activity — keeps the node alive and
+// syncing while the app is backgrounded, instead of being torn down with the UI
+// and killed by the OS.
+//
+// The Activity binds to this service to learn the loopback port() and to route
+// lifecycle "kicks" (onNetworkChanged / onResume) to the service-held App. The
+// ConnectivityManager.NetworkCallback also lives here (not in the Activity) so
+// the node re-dials on a network change even while backgrounded.
+class WalletService : Service() {
+
+    companion object {
+        private const val TAG = "bursa"
+
+        // Notification plumbing. The channel is created in onCreate (it MUST
+        // exist before startForeground posts a notification against it).
+        private const val CHANNEL_ID = "bursa_sync"
+        private const val NOTIFICATION_ID = 1
+
+        // Debounce window for network transitions (e.g. WiFi hand-off to
+        // cellular): a 500 ms window collapses back-to-back onLost→onAvailable
+        // events from a single interface change into one re-dial.
+        private const val DEBOUNCE_DELAY_MS = 500L
+    }
+
+    // The booted wallet. Null until started; guarded by walletLock because
+    // binder calls, connectivity callbacks, and service teardown can arrive on
+    // different threads. started gates double-start.
+    private val walletLock = Any()
+    private val walletExecutor = Executors.newSingleThreadExecutor { runnable ->
+        Thread(runnable, "bursa-wallet-lifecycle")
+    }
+    @Volatile
+    private var shuttingDown = false
+    private var reconnectInFlight = false
+    private var app: App? = null
+    private var started = false
+
+    // Set to the failure message when the node could not boot (app.start threw).
+    // The service keeps its binder answering so the Activity can read this and
+    // surface the failure instead of polling port() forever. Guarded by
+    // walletLock like the other lifecycle state.
+    private var bootError: String? = null
+
+    private val binder = LocalBinder()
+
+    private var connectivityManager: ConnectivityManager? = null
+    private var networkCallback: ConnectivityManager.NetworkCallback? = null
+
+    private val debounceHandler = Handler(Looper.getMainLooper())
+    private val reconnectRunnable = Runnable { handleNetworkChange() }
+
+    // LocalBinder is the in-process binding surface. The Activity casts the
+    // IBinder it receives in onServiceConnected to this and calls through to the
+    // service-held App. All methods are safe before the wallet has started
+    // (port() returns 0; the kicks no-op).
+    inner class LocalBinder : Binder() {
+        // port returns the loopback port the control surface bound, or 0 if the
+        // wallet has not finished starting. The Activity polls/loads only once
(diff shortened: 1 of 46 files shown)


mobile/android/app/src/main/java/io/blinklabs/bursa/WalletService.kt
```diff
@@ -0,0 +1,352 @@
...
+                    }
+                }
+            }
+        } catch (e: RejectedExecutionException) {
+            synchronized(walletLock) {
+                reconnectInFlight = false
+            }
+            android.util.Log.w(TAG, "$label rejected", e)
+        }
+    }
```
verbotenj, line 287: Holding the global monitor lock `walletLock` for the entire duration of the Go-side `action(instance)` blocks the Android main thread. Since blocking Go operations (like stopping/restarting the node) take several seconds to execute, any binder queries (`port()`, `bootError()`) or lifecycle callbacks (`onDestroy()`) running on the main UI thread will freeze and trigger an **Application Not Responding (ANR)** crash.

Extract the `app` instance quickly inside the synchronized block, then release the lock and perform the Go call safely on the background thread:

```suggestion
    private fun enqueueWalletKick(label: String, action: (App) -> Unit) {
        if (shuttingDown) return
        synchronized(walletLock) {
            if (shuttingDown || app == null || reconnectInFlight) return
            reconnectInFlight = true
        }
        try {
            walletExecutor.execute {
                var instance: App? = null
                synchronized(walletLock) {
                    if (!shuttingDown) {
                        instance = app
                    }
                }
                try {
                    instance?.let { action(it) }
                } catch (e: Exception) {
                    android.util.Log.w(TAG, "$label failed", e)
                } finally {
                    synchronized(walletLock) {
                        reconnectInFlight = false
                    }
                }
            }
        } catch (e: RejectedExecutionException) {
            synchronized(walletLock) {
                reconnectInFlight = false
            }
            android.util.Log.w(TAG, "$label rejected", e)
        }
    }
```

mobile/ios/Bursa/WalletViewController.swift
```diff
@@ -0,0 +1,238 @@
...
+    func onResume() {
+        let currentApp = syncState { () -> MobileApp? in
+            stopping ? nil : app
+        }
+        do {
+            try currentApp?.onResume()
+        } catch {
+            NSLog("bursa: wallet resume failed: \(error)")
+        }
+    }
```
verbotenj, line 152: **Severity:** 🟡 **High**

Executing the blocking Go-side `currentApp?.onResume()` call synchronously on the iOS main thread blocks the application during foregrounding. If this operation takes longer than the iOS watchdog threshold (typically 5 seconds), the system watchdog will abruptly terminate the application (generating a crash log with exception code `0x8badf00d` - "ate bad food") and freeze the UI.

Refactor `onResume()` to dispatch the blocking Go call asynchronously onto the background serial `stateQueue`:

```suggestion
    // onResume re-dials the node's peers after the app returns from the
    // background. Called by AppDelegate.applicationWillEnterForeground.
    func onResume() {
        stateQueue.async { [weak self] in
            guard let self = self else { return }
            let currentApp = self.syncState { () -> MobileApp? in
                self.stopping ? nil : self.app
            }
            do { 
                try currentApp?.onResume()
            } catch {
                NSLog("bursa: wallet resume failed: \(error)")
            }
        }
    }
```

ui/mobile/mobile.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+func cleanupLateStart(result <-chan startResult) {
+	go func() {
+		res := <-result
+		if res.app != nil {
+			_ = res.app.Stop()
+		}
+	}()
+}
```
verbotenj: **Severity:** 🟡 **High** (Performance/Reliability)

If the bootstrap process (`bootWallet`) gets stuck indefinitely (e.g. waiting on a database lock, unresponsive peer connection, or hung DNS resolution), the `result` channel will never receive a message. The background goroutine spawned by `cleanupLateStart` will block indefinitely on `<-result`, creating a permanent goroutine/memory leak.

Refactor `cleanupLateStart` to accept and select on context cancellation, ensuring the leaked goroutine terminates cleanly on timeout or startup failure:

```suggestion
func cleanupLateStart(ctx context.Context, result <-chan startResult) {
	go func() {
		select {
		case res := <-result:
			if res.app != nil {
				_ = res.app.Stop()
			}
		case <-ctx.Done():
			// Clean exit if the startup context was terminated due to stall/failure
		}
	}()
}
```

*(Note: Remember to also update the calls to `cleanupLateStart(result)` on lines 123 and 127 to pass `ctx` as well: `cleanupLateStart(ctx, result)`)*

mobile/ios/Bursa/WalletViewController.swift
```diff
@@ -0,0 +1,238 @@
...
+import Bursa
+
+// WalletViewController boots the in-process wallet (the embedded lean Dingo node
+// + the loopback control surface serving the API and the embedded SPA), then
+// points a full-screen WKWebView at the loopback URL the wallet chose.
+class WalletViewController: UIViewController, WKNavigationDelegate {
+
+    private var app: MobileApp?
+    private var webView: WKWebView!
+    private var walletPort = 0
```
verbotenj, line 38: **Severity:** 🟡 **Medium** (Security/Privacy)

Using legacy synchronous logging like `NSLog` writes raw, unredacted string interpolations (such as file paths, network settings, or error diagnostics) directly into the public system log console. This risks leaking private user metadata or sensitive operational details into public device logs.

Migrate to Apple's modern, lightweight, structured `OSLog` framework, which automatically enforces **privacy redactions** by default for string interpolations unless they are explicitly marked `.public`:

```suggestion
import UIKit
import WebKit
import OSLog

// Mobile is the gomobile-generated framework. `gomobile bind -target=ios`
// compiles the Go `mobile` package into Bursa.xcframework; the Go package name
// becomes the Objective-C symbol prefix `Mobile`. So the Go `App` struct is the
// class `MobileApp`, and the Go `New()` constructor is the free function
// `MobileNew()`. (Bursa.xcframework is the framework module Swift imports.)
import Bursa

// WalletViewController boots the in-process wallet (the embedded lean Dingo node
// + the loopback control surface serving the API and the embedded SPA), then
// points a full-screen WKWebView at the loopback URL the wallet chose.
class WalletViewController: UIViewController, WKNavigationDelegate {

    private let logger = Logger(subsystem: "io.blinklabs.bursa", category: "WalletViewController")
    private var app: MobileApp?
```

*(Note: After importing, replace legacy `NSLog` calls with the modern logger, keeping sensitive details redacted by default under Swift's automatic `.private` interpolation policies. For example, use `logger.error("bursa: wallet start failed: \(error.localizedDescription)")` without public markers to redact underlying exception fields from raw console inspection.)*

verbotenj (approved):

PR #563, opened 2026-07-03
feat(ui): ADA Handle resolution in Send

wolf31o2:
## Summary by cubic
Adds ADA Handle ($name) resolution to the Send flow using the embedded node. Users can enter a $handle and it resolves to the current holder’s address on supported networks with clear UI feedback.

- **New Features**
  - Send: recipient input accepts $handles; lookup is debounced, shows “Resolving/Resolved by your node/Handle not found”, disables Review until resolved, and builds the tx with the resolved address.
  - API: added GET /wallet/handle/{name} returning {handle, address}; case-insensitive, optional ‘$’; supports mainnet/preview/preprod (shared policy); 404 when not found or unsupported network, 503 when node unavailable or starting, 502 on node errors.

- **Refactors**
  - Added `ui/internal/handle` with normalization and `Resolve` (node-only lookup).
  - Extended `ui/internal/chain.Client` with `AssetAddresses` and `AssetAddress`.
  - Replaced `PoolDRepLookup` with `NodeLookup` in the API to support handle resolution.

<sup>Written for commit 39a51e0f45fe7be1590c313422ec9a9eec42c7e9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/bursa/pull/563?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ui/internal/api/api.go
+++ b/ui/internal/api/api.go
@@ -25,6 +25,7 @@ import (
 	"unicode/utf8"
 
 	"github.com/blinklabs-io/bursa/ui/internal/chain"
+	"github.com/blinklabs-io/bursa/ui/internal/handle"
 	"github.com/blinklabs-io/bursa/ui/internal/keystore"
 	"github.com/blinklabs-io/bursa/ui/internal/poolops"
 	"github.com/blinklabs-io/bursa/ui/internal/spend"
@@ -70,12 +71,15 @@ type Spender interface {
 	BuildDelegation(ctx context.Context, req spend.DelegationRequest) (spend.DelegationPreview, error)
 }
 
-// PoolDRepLookup is the node-backed verification surface the staking screen uses
-// to confirm a pasted pool or DRep ID exists (consent law: the embedded node is
-// the only thing that touches the network). *chain.Client satisfies it.
-type PoolDRepLookup interface {
+// NodeLookup is the node-backed verification surface the staking and send
+// screens use: confirming a pasted pool or DRep ID exists, and resolving an
+// ADA Handle ($name) to its current holding address (consent law: the
+// embedded node is the only thing that touches the network). *chain.Client
+// satisfies it.
+type NodeLookup interface {
 	Pool(ctx context.Context, poolID string) (chain.PoolInfo, error)
 	DRep(ctx context.Context, drepID string) (chain.DRepInfo, error)
+	AssetAddresses(ctx context.Context, asset string) ([]chain.AssetAddress, error)
 }
 
 // Vault is the encrypted multi-wallet store the API drives. It owns the wallet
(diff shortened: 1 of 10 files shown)


ui/internal/api/api.go
```diff
@@ -731,6 +744,21 @@ func NewHandler(st Statuser, vlt Vault, wl Wallet, sp Spender, settings Settings
 		serveLookup(w, info, err)
 	}))
 
+	// ADA Handle ($name) resolution for the Send screen: resolve the on-chain
+	// NFT to its current holding address through the node (never an external
+	// service). A leading '$' is optional; handle.Resolve normalizes it. A
+	// name with no Handle policy on this network (anything but mainnet, today)
+	// or with no on-chain holder reports as a clean not-found (404), never a
+	// hard error.
+	mux.HandleFunc("GET /wallet/handle/{name}", gated(st, func(w http.ResponseWriter, r *http.Request) {
```
verbotenj, line 752: 🟡 **UX Risk:** For non-mainnet networks (e.g., `preview`, `preprod`), `handle.Resolve` returns `chain.ErrNotFound` because no canonical Handle policy is defined. This causes the API to return a `404` and the frontend Send screen to show `"Handle not found"`. This is misleading because it implies the handle might be found if the name was different, when in fact handle resolution is completely unsupported on testnets.

**Proposed Fix:** Guard the endpoint and return a `400 Bad Request` with a clear explanation when the active network is not supported. Use `handle.PolicyForNetwork` to determine if resolution is available for the active network.

```go
	mux.HandleFunc("GET /wallet/handle/{name}", gated(st, func(w http.ResponseWriter, r *http.Request) {
		if _, ok := handle.PolicyForNetwork(network); !ok {
			writeJSON(w, http.StatusBadRequest, map[string]string{"error": "ADA Handle resolution is only available on mainnet"})
			return
		}
		if lookup == nil {
			writeJSON(w, http.StatusServiceUnavailable, map[string]string{"error": "handle lookup unavailable"})
			return
		}
		bare, addr, err := handle.Resolve(r.Context(), lookup, network, r.PathValue("name"))
		serveLookup(w, handleInfo{Handle: bare, Address: addr}, err)
	}))
```

wolf31o2: Per ADA Handle, the Policy ID is the same on all of the networks.

verbotenj (approved):

PR #619, opened 2026-07-23
feat(ui): Keystone hardware wallet (air-gapped QR + USB)

wolf31o2:
Adds Keystone as a fully **air-gapped QR** hardware signer on the `HardwareSigner` abstraction — account xpub via a `crypto-multi-accounts` sync QR, and tx signing via an animated-UR exchange (display the unsigned request → scan the signed witness back). Byte-identical xpub/witness output to Ledger/Trezor via the shared `hw/xpub.ts` + `hw/witness.ts` encoders.

## On-device fee display
The neutral backend hardware-sign request now carries each input's real **lovelace amount + address** (resolved at tx-build time from coin-selection — no extra node query), so the Keystone displays the actual inputs and **computes the fee**. On an air-gapped signer the on-device display *is* the security control.

## Transports
**QR is the only user-selectable transport.** A USB path (`@keystonehq/hw-app-ada`) exists in code but is intentionally **not user-selectable**, pending validation on real hardware.

## Consent-law
QR is fully offline (camera is local) — no network egress, so no external-consent gate.

## Capabilities
`{ send: true, staking: false, governance: false, multisig: false, poolReg: false }` — conservative; the QR request maps payment inputs + change today.

## Safety
Bounds-checked CBOR reader (throws on truncated frames; asserts 32-byte pubkey / 64-byte sig); a missing account fingerprint blocks signing with an account-sync re-scan prompt (never fabricates a zero xfp); unmatched/valueless inputs throw rather than silently drop.

## Verification
`npm run lint` (0 errors) · `npm run test` (579 passing) · `tsc --noEmit` · `vite build`; Keystone libs code-split out of the initial bundle. Go root + ui: build/vet/test green.

## Summary by cubic
Adds Keystone hardware wallet support with a local, air‑gapped QR flow. The QR sign request now includes real input values and addresses so the device can show inputs and compute the fee; USB remains gated and not user‑selectable.

- **New Features**
  - Air‑gapped QR end‑to‑end: import xpub from `crypto-multi-accounts`, show animated `cardano-sign-request`, scan `cardano-signature`, and build witnesses. New UI: `AnimatedQR`, `QRScanner`, `KeystoneQRModal` (+ `useKeystoneQRBridge`). Capabilities: send‑only.
  - On‑device input/fee display: backend retains each input’s lovelace, address, and assets and includes them in `HWInput`; Keystone maps them into its UTxOs for accurate on‑device review.
  - UI/Performance: Keystone is selectable as QR‑only; Add Wallet includes the account‑sync scan; Send uses the QR modal bridge. Keystone/USB libs are dynamically imported and code‑split with `vite-plugin-wasm` and `vite-plugin-top-level-await`. Local‑only; no external‑consent gate.

- **Bug Fixes**
  - Safety/hardening: block QR signing without a valid master fingerprint; persist/validate Keystone XFP (8 hex). Bounds‑check QR witness parsing and assert key sizes; error on missing paths/values and on empty UTxO sets.
  - QR UX: add Retry to remount the scanner; detect and surface corrupt multipart URs; dedicated centered overlay. Dependencies: add `@zxing/library`.
  - Transport: USB path kept in code but hidden from pickers and refused by `connectHardware` pending real‑device validation.

<sup>Written for commit 91d44026a9fd7f51a5be74eda017296ca98b43b8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/bursa/pull/619?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Screenshots

Rendered browser captures (Playwright, 1280×800) of the Keystone UI states reachable without a physical device.

**Add Wallet — device picker (Keystone enabled)**

Device picker with Keystone enabled

**Account-sync QR scan (xpub import)** — the webcam scanner modal; headless has no camera, so it shows the "Requested device not found" fallback.

Account-sync QR scan prompt

**Send — animated sign-request QR modal**

Send sign QR modal with animated UR

### DIFF
--- a/ui/web/package.json
+++ b/ui/web/package.json
@@ -13,8 +13,14 @@
   },
   "dependencies": {
     "@cardano-foundation/ledgerjs-hw-app-cardano": "^7.1.4",
+    "@keystonehq/bc-ur-registry-cardano": "^0.5.0",
+    "@keystonehq/hw-app-ada": "^0.1.1",
+    "@keystonehq/hw-transport-webusb": "^0.5.2",
     "@ledgerhq/hw-transport-webhid": "^6.35.4",
+    "@ngraveio/bc-ur": "^1.1.13",
     "@trezor/connect-web": "^9.7.3",
+    "@zxing/browser": "^0.1.5",
+    "@zxing/library": "^0.21.0",
     "bech32": "^1.1.4",
     "qrcode.react": "^4.2.0",
     "react": "^19.0.0",
@@ -33,6 +39,8 @@
     "typescript": "^5.5.0",
     "typescript-eslint": "^8.0.0",
     "vite": "^6.0.0",
+    "vite-plugin-top-level-await": "^1.6.0",
+    "vite-plugin-wasm": "^3.6.0",
     "vitest": "^3.0.0"
   }
 }
--- a/ui/web/src/hw/deviceKind.ts
+++ b/ui/web/src/hw/deviceKind.ts
@@ -26,10 +26,9 @@ import type { HardwareKind } from "./types";
 // read/write the SAME key as this module (one source of truth).
 export const STORAGE_KEY = "bursa.hw.deviceKind";
 
-// Kinds we accept from storage. Keystone is intentionally omitted: it is
-// disabled this phase, so a stale "keystone" hint must NOT select an
-// unsupported signer — it falls back to the documented "ledger" default.
-const KNOWN_KINDS: readonly HardwareKind[] = ["ledger", "trezor"];
+// Kinds we accept from storage. All implemented signers are listed; an
+// unrecognised value falls back to the documented "ledger" default.
+const KNOWN_KINDS: readonly HardwareKind[] = ["ledger", "trezor", "keystone"];
 
 type DeviceKindMap = Record<string, HardwareKind>;
 
@@ -92,4 +91,68 @@ export function getStoredDeviceKind(walletId: string): HardwareKind | undefined 
  */
 export function getDeviceKind(walletId: string): HardwareKind {
   return getStoredDeviceKind(walletId) ?? "ledger";
+}
+
+// ── Keystone master fingerprint (xfp) ────────────────────────────────────────
+//
+// A Keystone signing over QR must stamp its wallet master fingerprint on the
+// sign-request so the device recognises the witness paths as its own. That
+// fingerprint is learned only during account-sync (it rides the
+// crypto-multi-accounts QR), so it is remembered here — a purely local,
+// non-secret hint keyed by wallet id, exactly like the device-kind hint above.
+
+// localStorage key for the wallet-id → xfp map.
+export const KEYSTONE_XFP_KEY = "bursa.hw.keystoneXfp";
+
+function readXfpMap(): Record<string, string> {
+  if (typeof localStorage === "undefined") return {};
+  try {
+    const raw = localStorage.getItem(KEYSTONE_XFP_KEY);
+    if (!raw) return {};
+    const parsed = JSON.parse(raw) as unknown;
+    if (parsed && typeof parsed === "object") return parsed as Record<string, string>;
+    return {};
+  } catch {
+    return {};
+  }
+}
+
+/** Record the Keystone master fingerprint (hex) for a hardware wallet id. */
+export function setKeystoneXfp(walletId: string, xfp: string): void {
+  if (typeof localStorage === "undefined") return;
+  try {
+    const map = readXfpMap();
+    map[walletId] = xfp;
+    localStorage.setItem(KEYSTONE_XFP_KEY, JSON.stringify(map));
+  } catch {
+    // Best-effort: a full/blocked store just means the fingerprint is not
+    // remembered. The QR sign flow then treats the hint as absent and blocks
+    // signing (prompting an account-sync re-scan) rather than fabricating a zero.
+  }
+}
+
+/**
+ * A Cardano/BIP32 master fingerprint is exactly 4 bytes → 8 hex digits. Anything
+ * else in local storage is corrupt (e.g. a truncated write, a hand-edited value,
+ * or a non-string smuggled in by a malformed JSON blob) and MUST NOT be forwarded
+ * into a sign-request, where it would make the device fail to match its paths.
+ *
+ * "00000000" is deliberately VALID: a genuine all-zero fingerprint reported by a
+ * real device is a legitimate (1-in-2^32) value, and this guard blocks the
+ * ABSENCE of a fingerprint (undefined / corrupt storage), never a real one. The
+ * QR signer only refuses to FABRICATE a zero when the hint is missing (see the
+ * xfp note in keystone.ts) — it does not reject a device-reported zero.
+ */
+export function isValidKeystoneXfp(value: unknown): value is string {
+  return typeof value === "string" && /^[0-9a-f]{8}$/i.test(value);
+}
+
+/**
+ * Look up the stored Keystone master fingerprint, or `undefined` if unknown or
+ * malformed. Corrupt local state (a non-string, or a non-8-hex-digit value) is
+ * treated as unknown so it can never be forwarded into a QR signing request.
+ */
+export function getKeystoneXfp(walletId: string): string | undefined {
+  const xfp = readXfpMap()[walletId];
+  return isValidKeystoneXfp(xfp) ? xfp : undefined;
 }
--- a/ui/web/src/hw/keystone.ts
+++ b/ui/web/src/hw/keystone.ts
@@ -1,0 +1,647 @@
+/**
+ * hw/keystone.ts — Keystone hardware-wallet signer (air-gapped QR + USB).
+ *
+ * Keystone is fully LOCAL on BOTH transports, so — unlike the Trezor path — no
+ * external-consent gate applies (see the consent-law note in hw/types.ts):
+ *   - QR  (primary): fully offline. An animated QR carries a `cardano-sign-request`
+ *          UR to the device; the device signs on-screen and shows a
+ *          `cardano-signature` UR the user scans back through the webcam. Only
+ *          the local camera (getUserMedia) is touched — no network egress.
+ *   - USB (secondary, best-effort): @keystonehq/hw-app-ada over WebUSB. The
+ *          vendor SDK is young (v0.1.1) and firmware coverage is unverified, so
+ *          failures are surfaced plainly and capabilities stay conservative.
+ *
+ * Every heavy dependency (@keystonehq/*, buffer) is loaded through a dynamic
+ * import inside the connect path so it is CODE-SPLIT out of the initial bundle.
+ *
+ * Xpub/witness encoding parity with Ledger/Trezor/Go is guaranteed by reusing
+ * the SHARED hw/xpub.ts and hw/witness.ts encoders: the account-sync xpub and
+ * the assembled witness array are byte-for-byte identical for identical key
+ * material (exercised in keystone.test.ts).
+ */
+
+import type { HardwareSignResponse } from "../api/types";
+import type {
+  HardwareCapabilities,
+  HardwareSigner,
+  KeystoneConnectOptions,
+  KeystoneQRConnectOptions,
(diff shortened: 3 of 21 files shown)


ui/web/src/hw/keystone.ts
```diff
@@ -0,0 +1,567 @@
...
+          transactionHash: inp.tx_hash_hex,
+          index: inp.output_index,
+          // Per-input amount/address are not carried by the neutral request;
+          // they are display-only for the device (the signature is over
+          // signData), so empty/zero placeholders are used. See the human
+          // follow-up note in keystone.test.ts / the PR body.
+          amount: "0",
+          xfp,
+          hdPath: normalizePath(inp.path as string),
+          address: "",
```
chrisguiney: **Blocking.** Every wallet-owned input goes to the device with `amount: "0"` and `address: ""`.

The comment above defends this as display-only because the signature is over `signData`. That's correct about *what gets signed* — a bad placeholder can't cause a different transaction to be signed. But it's the wrong security argument for an **air-gapped** device: the on-device display *is* the control. The whole reason a user tolerates the QR dance is to confirm, on hardware the host cannot influence, what they're approving.

Concretely: a Cardano tx body carries outputs (address + amount) but inputs are only `(txid, index)`. Input amounts are exactly what this request must supply. Without them the Keystone cannot compute or display the **fee** — the one number a user most needs to sanity-check, and the one a compromised host would most want to inflate. The user ends up confirming a transaction the device can only partially describe.

The PR body already proposes the right fix (carry per-input amount/address on the neutral `HardwareSignResponse`). Until that lands, I'd either keep QR signing behind a flag, or show an explicit in-app warning that on-device input details are placeholders and the user must verify amount and fee in the app.

ui/web/src/screens/AddWallet.tsx
```diff
@@ -498,6 +539,36 @@ export function AddWallet({
...
+                <input
+                  type="radio"
+                  name="hw-keystone-transport"
+                  value="usb"
+                  checked={keystoneTransport === "usb"}
+                  disabled={loading}
+                  onChange={() => { setError(null); setKeystoneTransport("usb"); }}
+                  aria-label="USB"
+                />
+                USB cable
```
chrisguiney: **Blocking.** The PR description is candid that USB is "best-effort", rides a "young, unverified SDK" (`@keystonehq/hw-app-ada@^0.1.1`, pre-1.0), and that "firmware coverage is unverified — should be tested against a real device."

None of that reaches the user. I grepped `AddWallet.tsx` and `Send.tsx` for `experimental|best-effort|unverified|beta` — zero hits. Someone picking "USB cable" here gets no signal that this signing path has never been exercised against real hardware.

For a wallet, an untested signing transport should either be labeled in the UI or not offered yet. A parenthetical on this label plus a line of helper text would be enough.

ui/web/src/hw/keystone.ts
```diff
@@ -0,0 +1,567 @@
...
+// but the backend's submit endpoint wants the RAW vkey-witness array
+// ([[pubkey, sig], …]) that hw/witness.ts produces. We therefore decode just
+// enough CBOR to pull the vkey witnesses (map key 0) out of the witness set and
+// re-encode them through the shared encoder, so Keystone's witness output is
+// byte-identical to Ledger's and Trezor's.
+
+type CborValue = number | Uint8Array | string | CborValue[] | Map<number, CborValue>;
+
+/** Decode one CBOR data item at `offset`; returns the value and the next offset. */
+function decodeCbor(buf: Uint8Array, offset: number): { value: CborValue; next: number } {
```
chrisguiney, line 136: This decoder parses input that arrives **off a camera** — reachable if a user is induced to scan a hostile QR — and it does no bounds checking. Past the end of `buf`, JS yields `undefined`:

- `info === 24`: `len = buf[pos]` -> `undefined` -> `buf.slice(pos, NaN)` -> **empty array, no error**
- `info === 25`: `(undefined << 8) | undefined` -> `0` -> silently zero-length

`witnessSetToPairs` then only checks `instanceof Uint8Array`, not length, so a truncated frame can produce a zero-length signature that flows into `encodeWitnessArray`. The result fails at submission — fail-closed, so this is confusion/DoS rather than fund loss — but it should throw at the parse rather than yield a plausible-looking empty witness.

Suggest a `pos + n <= buf.length` guard before each multi-byte read, plus asserting `pub.length === 32` and `sig.length === 64` in `witnessSetToPairs`.

Separately, worth asking before keeping a bespoke parser: `@ngraveio/bc-ur` and `@keystonehq/bc-ur-registry-cardano` are both imported in `signTx` already and both do CBOR internally. Does `CardanoSignature` expose the parsed witnesses directly? There's no pre-existing CBOR decoder in this repo, so this isn't duplicating repo code — but reimplementing one when two in-scope dependencies have one is new attack surface in a wallet.

ui/web/src/hw/deviceKind.ts
```diff
@@ -93,3 +92,60 @@ export function getStoredDeviceKind(walletId: string): HardwareKind | undefined
...
+  }
+}
+
+/**
+ * A Cardano/BIP32 master fingerprint is exactly 4 bytes → 8 hex digits. Anything
+ * else in local storage is corrupt (e.g. a truncated write, a hand-edited value,
+ * or a non-string smuggled in by a malformed JSON blob) and MUST NOT be forwarded
+ * into a sign-request, where it would make the device fail to match its paths.
+ */
+export function isValidKeystoneXfp(value: unknown): value is string {
```
chrisguiney, line 146: `/^[0-9a-f]{8}$/i` accepts `"00000000"`, and `deviceKind.test.ts:61` **deliberately asserts** it's valid. Meanwhile `keystone.ts:356` says *"Never fall back to a zero fingerprint"* and `keystone.test.ts:164` says *"must block rather than silently send 00000000"*.

I read the intent as: block *absence* so the code never **fabricates** a zero, while still accepting a genuine (1-in-2^32) all-zero fingerprint reported by a real device. If that's right, the test is correct and the prose is just loosely worded — but as written the two contradict, and a future reader could "fix" either side in the wrong direction.

Suggest rewording the `keystone.ts` comment to "never *fabricate* a zero fingerprint" and adding a line here explaining why all-zeros is intentionally valid.

ui/web/src/hw/keystone.ts
```diff
@@ -0,0 +1,567 @@
...
+        );
+      }
+      // Map the neutral request → a Keystone CardanoSignRequest:
+      //   signData    = the unsigned tx body the device signs;
+      //   utxos       = the wallet-owned inputs (path → which key witnesses);
+      //   extraSigners= none (send + simple witnessing only; structured
+      //                 multisig/cert signers are gated off by capabilities).
+      const signData = B.from(req.unsigned_tx_cbor, "hex") as unknown as Buffer;
+      const utxos = req.inputs
+        .filter((inp) => inp.path)
```
chrisguiney: An input with no derivation path is silently dropped from `utxos` with no diagnostic. For send-only wallets every input should be owned, so this is inert today — but if it ever isn't, the device receives an incomplete UTxO list and the returned witness set won't cover all required keys. That fails closed, but opaquely.

An explicit throw naming the unmatched input, plus a guard for `utxos.length === 0` before constructing the request, turns a silent wrong-shape request into a clear error.

ui/web/src/hw/keystone.ts
```diff
@@ -0,0 +1,567 @@
...
+// ── Capabilities ───────────────────────────────────────────────────────────
+
+// QR is a "sign a raw tx body + witness these paths" model. Today only the
+// payment inputs are mapped into the request's `utxos`; stake-key signer paths
+// are NOT yet mapped into `extraSigners`, so a certificate tx would be witnessed
+// incompletely. Staking is therefore advertised as UNSUPPORTED until those
+// signer paths are wired in — don't flip this back to `true` without also
+// populating `extraSigners` with the stake path(s). Structured multisig /
+// pool-reg / governance are likewise not expressible here and stay gated off.
+const KEYSTONE_QR_CAPABILITIES: HardwareCapabilities = {
```
chrisguiney, line 73: The code is right (`staking: false`), but the PR description's hand-written "Capabilities" section claims QR is `{ send: true, staking: true, ... }`. The cubic summary agrees with the code ("QR is send-only; staking disabled until stake paths are mapped"), so the hand-written section looks stale from an earlier commit.

Worth fixing in the description — a reviewer reading only the body would believe staking is supported and wouldn't go looking for the gate.

ui/web/package.json
```diff
@@ -13,8 +13,14 @@
   },
   "dependencies": {
     "@cardano-foundation/ledgerjs-hw-app-cardano": "^7.1.4",
+    "@keystonehq/bc-ur-registry-cardano": "^0.5.0",
+    "@keystonehq/hw-app-ada": "^0.1.1",
+    "@keystonehq/hw-transport-webusb": "^0.5.2",
     "@ledgerhq/hw-transport-webhid": "^6.35.4",
+    "@ngraveio/bc-ur": "^1.1.13",
     "@trezor/connect-web": "^9.7.3",
+    "@zxing/browser": "^0.1.5",
```
chrisguiney, line 22: Five new runtime dependencies, three of them 0.x (`@keystonehq/bc-ur-registry-cardano@^0.5.0`, `@keystonehq/hw-app-ada@^0.1.1`, `@zxing/browser@^0.1.5`). `^` on a 0.x range is minor-locked by npm semantics, so the ranges themselves are fine.

Raising it as a conscious call rather than an implicit one: this is a meaningful supply-chain expansion for a wallet, and `@zxing/*` gets camera access. The code-splitting work is the right mitigation for load cost but not for supply-chain risk.

chrisguiney (changes requested):
Requesting changes. The engineering here is careful and the shared-helper reuse is exactly right — `encodeXpub`/`encodeWitnessArray` are used on **both** the QR and USB paths, which is what actually delivers the claimed byte-parity with Ledger and Trezor rather than a fourth bespoke encoder. My objections are about what gets enabled for users of a wallet, not about code quality.

Two blocking items, both already disclosed in the PR description — my concern is that they ship enabled anyway, and a note in a PR body isn't a mitigation for the person holding the keys:

1. **Air-gapped QR signing sends placeholder input data**, so the device cannot display or compute the fee. On an air-gapped signer the on-device display *is* the security control; "display-only" is the reason it matters, not a reason to skip it.
2. **The USB path is user-selectable with no on-device verification and no UI labeling**, despite the description calling it "best-effort" on a "young, unverified SDK".

Five non-blocking comments inline as well.

## Verification performed

| Check | Result |
|---|---|
| CI on this PR | **all green** — `web` (JS tests), `ui-lint`, `ui-go`, `extension`, CodeQL (incl. javascript-typescript), nilaway, DCO |
| Local `npm test` / `lint` / `type-check` / `build` | **not run — no Node.js on my machine** |

Flagging that plainly: I could not independently verify the 544 passing tests or the bundle-splitting claim, and everything below comes from reading the diff. I'm relying on CI for the test result.

No linked issue, and `gh issue list --search keystone` returns nothing, so there's no spec to check the implementation against beyond the description.

## Duplication

Clean, and deliberately so — see the note above about `xpub.ts`/`witness.ts` reuse. The one place I'd push back is the hand-rolled CBOR reader (inline comment): there's no pre-existing CBOR decoder in the repo, so it isn't duplicating repo code, but two dependencies already imported in that same function do CBOR internally.

## What's good

The `HardwareSigner` abstraction absorbs a genuinely awkward transport — a two-way animated-QR exchange with a human in the loop — without leaking into `Send`/`AddWallet` beyond a modal bridge. Capabilities are conservative on both transports. The xfp-missing path gives an actionable recovery instruction naming the exact device menu instead of a generic failure. And the extended-vkey trim in `witnessSetToPairs` (64-byte `pubkey||chaincode` to the leading 32 bytes) is the kind of firmware-variance detail that usually surfaces as a bug report months later.

chrisguiney (approved):

PR #620, opened 2026-07-23
fix: []byte mnemonic derivation to avoid un-zeroable string in signing paths

### LINKED ISSUE #591: Signing paths copy the mnemonic into un-zeroable Go strings
## Context

Surfaced during review of #554 (CIP-30 dApp connector):
https://github.com/blinklabs-io/bursa/pull/554#pullrequestreview-4718326140

## Problem

Every signing path derives the root key with:

```go
rootKey, err := bursa.GetRootKeyFromMnemonic(string(mnemonicBytes), "")
```

`string(mnemonicBytes)` allocates an **immutable** Go string containing the decrypted mnemonic.
Because strings are immutable, this copy cannot be zeroed — it lingers on the heap until GC reclaims
it (and may be relocated/copied again by the runtime). The surrounding code carefully zeroes
`mnemonicBytes` and the derived `bip32.XPrv` material (see `ui/internal/spend/spend.go`), but the
transient string defeats part of that effort.

Call sites include (non-exhaustive): `ui/internal/spend/spend.go:441, 499, 555, 859, 1166, 1428`.

## Suggested fix

Provide a `[]byte`-based derivation entry point (e.g. `bursa.GetRootKeyFromMnemonicBytes([]byte)`),
or have `GetRootKeyFromMnemonic` accept a byte slice / zeroable buffer, so no immutable string copy
of the mnemonic is created. Update the signing paths to use it and zero the buffer as they already do.

## Severity

Best-effort hardening (defence-in-depth for in-memory secrets), consistent with the zeroization
already performed on these paths. Not an active exploit.

wolf31o2:
Closes #591.

Every signing path derived the root key with `GetRootKeyFromMnemonic(string(mnemonicBytes), "")`. The `string()` conversion allocates an immutable copy of the decrypted mnemonic that cannot be zeroed and lingers on the heap, defeating the surrounding zeroing of `mnemonicBytes`.

## Root module
- Add `GetRootKeyFromMnemonicBytes(mnemonic []byte, password string) (bip32.XPrv, error)`, deriving the root key from a zeroable byte slice with no intermediate mnemonic string.
- go-bip39 v0.2.0 has no byte-oriented decode — `IsMnemonicValid`/`EntropyFromMnemonic` both take a `string` — so the BIP39 mnemonic→entropy decode (word lookup + checksum) is reimplemented over bytes using only its word list.
- `GetRootKeyFromMnemonic` now delegates to the byte version.

## ui module
- The signing paths (`spend`, `multisig`, `poolops`) now derive via the `wallet` package's byte-based `RootKeyFromMnemonicBytes` (the existing `AccountXpub` byte decoder, now exported), so the decrypted mnemonic is never converted to a string. Existing `mnemonicBytes` zeroing is preserved.
- The ui module pins bursa to a published commit (no local replace directive), so it routes through its own byte decoder rather than the newly added root helper.

## Tests
- Root: `GetRootKeyFromMnemonicBytes` derives the same root key as the string version for the canonical `abandon … about` mnemonic, and rejects invalid input.
- ui: `wallet.RootKeyFromMnemonicBytes` matches the string derivation and rejects invalid input.

## Verification
- Root: `CGO_ENABLED=0 go build ./...`, `CGO_ENABLED=0 go test ./...`, `golangci-lint run ./...` — all pass.
- ui: `go build/vet/test ./cmd/... ./internal/...` and `golangci-lint` on touched packages — all pass.

---
## Summary by cubic
Derives CIP-1852 root keys from zeroable mnemonic byte slices to avoid immutable strings in signing paths, fixing #591. Standardizes invalid-mnemonic handling and validates the byte decoder against upstream `go-bip39`.

- **Bug Fixes**
  - Added `GetRootKeyFromMnemonicBytes([]byte, string)` in `bursa` with a byte-only BIP39 decode; `GetRootKeyFromMnemonic` now delegates, zeroing transient entropy/password bytes and returning `ErrInvalidMnemonic` uniformly.
  - Exported `wallet.RootKeyFromMnemonicBytes([]byte)` and updated `ui` signing paths (`spend`, `multisig`, `poolops`) and account derivations to use it, removing `string(mnemonicBytes)` conversions.
  - Added tests in `bursa` and `ui/wallet` that validate the byte decoder and derived root keys against upstream `go-bip39` for 12–24 words (with/without password) and reject invalid mnemonics.

<sup>Written for commit 042fb406a6cd6ed64dfcaabcd656d2b6246dca8c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/bursa/pull/620?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added support for deriving wallet root keys directly from mnemonic byte data.
  - Improved handling of mnemonic data by clearing sensitive temporary values after use.
  - Standardized invalid mnemonic errors for more consistent behavior.

- **Bug Fixes**
  - Wallet, spending, pool, and multisignature operations now use the updated mnemonic processing across supported key-derivation flows.

- **Tests**
  - Added coverage for valid 12-, 15-, 18-, 21-, and 24-word mnemonics, passwords, and invalid inputs.

### DIFF
--- a/bursa.go
+++ b/bursa.go
@@ -619,20 +619,19 @@ func ExtractKeyFiles(wallet *Wallet) (map[string]string, error) {
 	return result, nil
 }
 
+// GetRootKeyFromMnemonic derives the CIP-1852 root key from a mnemonic string.
+// It delegates to GetRootKeyFromMnemonicBytes; callers that hold the decrypted
+// mnemonic in a zeroable byte slice (e.g. signing paths) should call that
+// function directly to avoid materializing an un-zeroable mnemonic string.
+//
+// Behavior change: every invalid mnemonic now returns ErrInvalidMnemonic
+// uniformly. Previously the underlying bip39.EntropyFromMnemonic error could be
+// surfaced; callers matching on that specific error should match on
+// ErrInvalidMnemonic instead.
 func GetRootKeyFromMnemonic(mnemonic, password string) (bip32.XPrv, error) {
-	if !bip39.IsMnemonicValid(mnemonic) {
-		return nil, ErrInvalidMnemonic
-	}
-	entropy, err := bip39.EntropyFromMnemonic(mnemonic)
-	if err != nil {
-		return nil, err
-	}
-	pwBytes := []byte{}
-	if password != "" {
-		pwBytes = []byte(password)
-	}
-	rootKey := GetRootKey(entropy, pwBytes)
-	return rootKey, nil
+	b := []byte(mnemonic)
+	defer zeroBytes(b)
+	return GetRootKeyFromMnemonicBytes(b, password)
 }
 
 func GetRootKey(entropy []byte, password []byte) bip32.XPrv {
--- a/bursa_test.go
+++ b/bursa_test.go
@@ -584,6 +584,79 @@ func TestGetRootKeyFromMnemonic(t *testing.T) {
 	}
 }
 
+func TestGetRootKeyFromMnemonicBytes(t *testing.T) {
+	// Independent oracle. The byte decoder is validated against
+	// go-bip39's EntropyFromMnemonic (the upstream decoder, not the code
+	// under test) rather than against GetRootKeyFromMnemonic — which now
+	// delegates to GetRootKeyFromMnemonicBytes and so would only compare the
+	// decoder with itself. For each of the five legal BIP39 word counts a valid
+	// mnemonic is sourced from fixed entropy via go-bip39's encoder, decoded
+	// back through the upstream decoder, and both entropyFromMnemonicBytes and
+	// GetRootKeyFromMnemonicBytes are asserted to agree. This exercises every
+	// checksumShift value (12/15/18/21/24 words -> shift 4/3/2/1/0).
+	t.Run("matches upstream decoder", func(t *testing.T) {
+		for _, words := range []int{12, 15, 18, 21, 24} {
+			// entropyLen bytes: 4 per 3 words (128..256 bits).
+			entropy := make([]byte, words/3*4)
+			for i := range entropy {
+				entropy[i] = byte(i*7 + 3) // deterministic, non-uniform
+			}
+
+			// Source a valid mnemonic from fixed entropy via the upstream
+			// encoder, so the vector is a real BIP39 mnemonic of this length.
+			mnemonic, err := bip39.NewMnemonic(entropy)
+			require.NoError(t, err, "%d words: NewMnemonic", words)
+			require.Len(t, strings.Fields(mnemonic), words)
+
+			// Oracle: decode with the upstream decoder, independent of the
+			// code under test.
+			wantEntropy, err := bip39.EntropyFromMnemonic(mnemonic)
+			require.NoError(t, err, "%d words: EntropyFromMnemonic", words)
+			assert.Equal(t, entropy, wantEntropy,
+				"%d words: upstream round-trip", words)
+
+			gotEntropy, err := entropyFromMnemonicBytes([]byte(mnemonic))
+			require.NoError(t, err, "%d words: entropyFromMnemonicBytes", words)
+			assert.Equal(t, wantEntropy, gotEntropy,
+				"%d words: byte decoder disagrees with upstream", words)
+
+			for _, password := range []string{"", "testpassword"} {
+				var pwBytes []byte
+				if password != "" {
+					pwBytes = []byte(password)
+				}
+				want := GetRootKey(wantEntropy, pwBytes)
+				got, err := GetRootKeyFromMnemonicBytes([]byte(mnemonic), password)
+				require.NoError(t, err,
+					"%d words: GetRootKeyFromMnemonicBytes(%q)", words, password)
+				assert.Equal(t, []byte(want), []byte(got),
+					"%d words: root key mismatch for password %q", words, password)
+			}
+		}
+	})
+
+	t.Run("rejects invalid mnemonic", func(t *testing.T) {
+		// A 12-word mnemonic whose last token is not in the BIP39 word list:
+		// legal length, so it reaches word lookup and fails there.
+		unknownWord := "abandon abandon abandon abandon abandon abandon " +
+			"abandon abandon abandon abandon abandon zzzzzz"
+		// All-"abandon": every token is a valid word and the length is legal,
+		// so it reaches checksum verification and fails there (the valid
+		// all-zero-entropy mnemonic ends in "about", not "abandon"). Confirmed
+		// checksum-invalid via bip39.EntropyFromMnemonic (returns an error).
+		_, err := bip39.EntropyFromMnemonic(
+			"abandon abandon abandon abandon abandon abandon " +
+				"abandon abandon abandon abandon abandon abandon")
+		require.Error(t, err, "sanity: all-abandon must be checksum-invalid")
+		invalidChecksum := "abandon abandon abandon abandon abandon abandon " +
+			"abandon abandon abandon abandon abandon abandon"
+		for _, m := range []string{"", "invalid mnemonic", unknownWord, invalidChecksum} {
+			_, err := GetRootKeyFromMnemonicBytes([]byte(m), "")
+			assert.ErrorIs(t, err, ErrInvalidMnemonic)
+		}
+	})
+}
+
 func TestCIP0003KeyGeneration(t *testing.T) {
 	// Test CIP-0003 compliance for wallet key generation
 	// These test vectors validate the complete key generation pipeline:
--- a/mnemonic_bytes.go
+++ b/mnemonic_bytes.go
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//	http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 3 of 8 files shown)


bursa_test.go
```diff
@@ -584,6 +584,32 @@ func TestGetRootKeyFromMnemonic(t *testing.T) {
 	}
 }
 
+func TestGetRootKeyFromMnemonicBytes(t *testing.T) {
+	const canonical = "abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about"
+
+	t.Run("matches string version", func(t *testing.T) {
```
chrisguiney: **Blocking.** This subtest is circular and cannot fail.

`want` comes from `GetRootKeyFromMnemonic(canonical, password)`, but this PR rewrote that function (`bursa.go:626`) to be:

```go
return GetRootKeyFromMnemonicBytes([]byte(mnemonic), password)
```

So `want` and `got` now run the *same* code. The assertion is `f(x) == f(x)`. If `entropyFromMnemonicBytes` produced entirely wrong entropy, both sides would be wrong identically and this would still pass — meaning nothing in the root module guards the new decoder at all.

Fix: assert against a **fixed known-good root key** (hex literal), or against `bip39.EntropyFromMnemonic` directly. Given this is key derivation in a wallet, I'd want it in this PR rather than as a follow-up.

Related trap worth heading off: the ui test (`ui/internal/wallet/mnemonic_bytes_test.go`) *is* a genuine cross-check today, but only because the ui module pins bursa to `v0.16.1-0.20260723122856-2f8b123a98ae` — a pre-this-PR commit whose `GetRootKeyFromMnemonic` still calls `bip39.EntropyFromMnemonic`. The moment ui bumps bursa past this PR, that test silently becomes circular too. Both should assert fixed vectors.

bursa.go
```diff
@@ -619,20 +619,12 @@ func ExtractKeyFiles(wallet *Wallet) (map[string]string, error) {
 	return result, nil
 }
 
+// GetRootKeyFromMnemonic derives the CIP-1852 root key from a mnemonic string.
+// It delegates to GetRootKeyFromMnemonicBytes; callers that hold the decrypted
+// mnemonic in a zeroable byte slice (e.g. signing paths) should call that
+// function directly to avoid materializing an un-zeroable mnemonic string.
 func GetRootKeyFromMnemonic(mnemonic, password string) (bip32.XPrv, error) {
```
chrisguiney, line 631: Two things on this wrapper:

**1. It leaves an un-zeroed byte copy of the mnemonic.** `[]byte(mnemonic)` allocates a fresh copy that nothing ever zeroes, and the caller has no handle on it. So the legacy string path now leaves the un-zeroable string *plus* an un-zeroed heap copy of the same secret — strictly more residue than before, on the path this PR is hardening. Two-line fix:

```go
b := []byte(mnemonic)
defer zeroBytes(b)
return GetRootKeyFromMnemonicBytes(b, password)
```

**2. Public error behavior changed.** This previously could surface `bip39.EntropyFromMnemonic`'s own error; every failure now collapses to `ErrInvalidMnemonic`. Arguably an improvement (uniform, and avoids leaking parse detail), but it's a behavior change on an exported function and isn't mentioned in the description — worth a line for anyone matching on the old error.

mnemonic_bytes.go
```diff
@@ -0,0 +1,128 @@
...
+	pwBytes := []byte{}
+	if password != "" {
+		pwBytes = []byte(password)
+	}
+	return GetRootKey(entropy, pwBytes), nil
+}
+
+// entropyFromMnemonicBytes decodes BIP39 entropy from a mnemonic byte slice,
+// verifying the checksum, without allocating an immutable mnemonic string.
+func entropyFromMnemonicBytes(mnemonic []byte) ([]byte, error) {
```
chrisguiney, line 56: Only 12-word mnemonics are tested. Both new tests use a single 12-word vector (`abandon … about`), so the 15/18/21/24-word paths — and specifically the `checksumShift` values 3/2/1/0 — are entirely unexercised.

I hand-checked the arithmetic for all five lengths and it's correct (table in the review body), but that should be a test rather than a reviewer's arithmetic, especially for a hand-rolled decoder. The BIP39 English reference vectors cover every length and are cheap to add here.

ui/internal/wallet/mnemonic_bytes.go
```diff
@@ -23,10 +23,18 @@ import (
...
+// full mnemonic as an immutable Go string, so signing paths that hold the
+// decrypted mnemonic in a zeroable buffer can derive without leaving an
+// un-zeroable copy on the heap. The caller keeps ownership of, and must zero,
+// the mnemonic slice it passes in.
+//
+// The ui module pins the bursa root module to a published commit (there is no
+// local replace directive), so this lives here rather than calling a root-module
+// helper: it lets the wallet's own packages derive from bytes without depending
+// on an unreleased bursa version.
+func RootKeyFromMnemonicBytes(mnemonic []byte) (bip32.XPrv, error) {
```
chrisguiney, line 37: This file and the root module's `mnemonic_bytes.go` now contain two copies of the same hand-rolled BIP39 decode. I diffed `entropyFromMnemonicBytes`, `writeMnemonicIndexBits`, `mnemonicWordIndex` and `zeroBytes` between them: **byte-identical apart from doc comments and the `bursa.` error qualifier** — roughly 65 lines of security-critical crypto duplicated.

The doc comment explains why (ui pins a published bursa commit with no `replace` directive), and that's a legitimate merge-ordering constraint, not a design choice. But two independent copies of key-derivation code will drift, and a bug fixed in one won't be fixed in the other.

Once this merges and bursa is tagged, ui can bump and delete this copy in favour of `bursa.GetRootKeyFromMnemonicBytes`. Worth filing that follow-up now so the duplication doesn't quietly become permanent.

chrisguiney (changes requested):
Requesting changes on one item: the only test guarding the new key-derivation code cannot fail. Everything else below is a comment.

The change delivers exactly what #591 asked for — I grepped for `GetRootKeyFromMnemonic(string(` and `string(mnemonicBytes)` across `ui/` and the root module and there are **zero remaining**, so all six call sites the issue lists are converted.

## Verification performed

| Check | Result |
|---|---|
| Root: `CGO_ENABLED=0 go build ./...` / `go test ./...` | pass (all packages) |
| Root: `golangci-lint run ./...` | 0 issues |
| ui: `CGO_ENABLED=0 go build ./...` | clean |
| ui: `go test ./internal/{wallet,spend,multisig,poolops}/...` | pass |
| ui: `golangci-lint` on touched packages | 0 issues |
| CI on this PR | all green |

## The BIP39 reimplementation is correct

Since this is hand-rolled key derivation I checked the arithmetic rather than trusting it. For all five legal word counts:

| Words | `entropyLen` | `checksumBits` | buffer bits | needed | `checksumShift` |
|---|---|---|---|---|---|
| 12 | 16 B (128) | 4 | 136 | 132 | 4 |
| 15 | 20 B (160) | 5 | 168 | 165 | 3 |
| 18 | 24 B (192) | 6 | 200 | 198 | 2 |
| 21 | 28 B (224) | 7 | 232 | 231 | 1 |
| 24 | 32 B (256) | 8 | 264 | 264 | 0 |

All correct. `writeMnemonicIndexBits` packs big-endian 11-bit groups correctly and the checksum compares `data[entropyLen] >> shift` against `sha256(entropy)[0] >> shift`.

Two details I went looking for and found done right:

- **`bytes.Fields` matches upstream.** go-bip39 v0.2.0's `EntropyFromMnemonic` routes through `splitMnemonicWords`, which uses `strings.Fields` and the identical `n%3 != 0 || n < 12 || n > 24` guard. This is a faithful port, not a guess — no divergence in what's accepted. (`MnemonicToByteArray` uses `strings.Split(mnemonic, " ")`, but that's a different entry point and isn't what's being replaced.)
- **No hidden string materialization.** `bytes.Fields` returns subslices of the caller's buffer, so zeroing it zeroes the words too; and `mnemonicWordIndex` compares `word[j] != candidate[j]` against a `string` candidate, and indexing a string yields a byte with no allocation. The mnemonic genuinely never becomes a Go string. `entropy` and `data` are both `defer zeroBytes`'d.

## Worth noting, not fixing

Go's GC may relocate a `[]byte` before you zero it, so `zeroBytes` clears the *current* backing array only. The issue already frames this as best-effort defence-in-depth and this PR meaningfully improves the situation — raising it only so the guarantee isn't over-stated later. The docstring's "callers remain responsible for zeroing the byte slice they pass in" is the right framing.

## What's good

The scope discipline is right: one narrow entry point, the string version kept as a delegating wrapper so nothing else has to change, and the existing `mnemonicBytes` zeroing in the signing paths left intact. Porting `splitMnemonicWords`' exact semantics rather than reaching for the more obvious `strings.Split(" ")` is the kind of detail that decides whether a "faithful reimplementation" actually is one. And reimplementing only the decode — while still deriving through the existing `GetRootKey` — keeps the actual key-stretching path untouched.

chrisguiney (approved):

PR #646, opened 2026-07-31
fix(ui): retry tx build when apollo hits the min-UTxO dust-change dead-zone

wolf31o2:
Restores the pre-upstream-apollo behavior in the send path's `Build` when a send would leave change just below the min-UTxO floor.

- Extracts the send build loop into `completeSend`, parameterized by a `forceInputs` flag. With `forceInputs` false, Apollo selects inputs from the loaded pool; with it true, every loaded UTxO is added as an explicit input (`AddInput`) and no coin selection runs.
- Detects Apollo's non-convergence from `Complete()` via `isNonConvergenceError` (matches `"evaluation transaction did not converge"`).
- On non-convergence, retries the build once with `forceInputs=true` so the combined change clears the min-UTxO floor.
- If the retry still cannot converge, returns `ErrInsufficientFunds` wrapped with "the amount leaves change below the minimum UTxO and no additional inputs are available to cover it" instead of the raw Apollo iteration message. Insufficient-funds errors from the retry keep mapping to `ErrInsufficientFunds`.
- The normal (non-dead-zone) path is unchanged.

Tests (`spend_test.go`):
- `TestBuildRetriesDustChangeDeadZone`: two ~5-ADA UTxOs, ~4-ADA send. Asserts `Build` succeeds, includes both inputs, has non-zero fee and non-zero change, and the tx signs and submits via `Confirm`.
- `TestBuildDustChangeDeadZoneNoExtraInputs`: single ~5-ADA UTxO, ~4-ADA send. Asserts `ErrInsufficientFunds` with the actionable message and no leaked "did not converge" text.

---
## Summary by cubic
Fixes tx building when `apollo` oscillates in the min-UTxO dust-change dead-zone by retrying with the smallest-that-works, largest-first input prefix. Adds a signed-size guard that now protects both retry and normal builds to prevent oversized transactions before approval.

- **Bug Fixes**
  - Detect `Complete()` non-convergence and retry with a growing, largest-first forced-input prefix; stop at the minimum that converges; return `ErrInsufficientFunds` with a clear dust-change message when no extra inputs exist.
  - Guard signed size using `backend` protocol params: project CBOR + one vkey witness per distinct input address, add witness-set overhead only when the set is empty, and reject over-limit with `ErrInsufficientFunds`; applies to coin-selected and forced-input builds.
  - Map remaining errors via `mapCompleteErr(...)`, honor canceled contexts, and avoid leaking raw `apollo` text.

- **Refactors**
  - Extracted `completeSend(...)`, `completeForcingPrefix(...)`, `guardTxSize(...)`, and `mapCompleteErr(...)`.
  - Capped signer-convergence attempts at min(loaded UTxOs, derived receive addresses) + 2.

<sup>Written for commit f245f70ab28f9f36fdc40f2478cec632d5f82c6f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/bursa/pull/646?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction building for low-value change scenarios, retrying with additional inputs when needed.
  * Returned clearer insufficient-funds errors when a transaction can’t be completed.
  * Reduced the chance of oversized transactions by limiting input selection to what can still be signed and submitted.

* **Tests**
  * Added coverage for retry behavior, insufficient-funds handling, and bounded input selection on large wallets.

### DIFF
--- a/ui/internal/spend/spend.go
+++ b/ui/internal/spend/spend.go
@@ -504,8 +504,16 @@ func (s *Service) completeSend(
 // array(2) header (1) + bytes(32) (2+32) + bytes(64) (2+64) = 101 bytes, rounded
 // up to 102 to cover the per-element set overhead. Complete() leaves the witness
 // set empty (witnesses are attached later at Confirm/Sign time), so the size
-// guard projects the SIGNED size by adding one such witness per input.
+// guard projects the SIGNED size by adding one such witness per distinct signer.
 const vkeyWitnessSizeEstimate = 102
+
+// witnessSetOverhead covers the one-time CBOR framing that appears when the
+// empty witness set gains its first vkey-witness field: the witness-set map key
+// plus the vkey-witness array header. It is added once (not per witness), so a
+// transaction sitting a few bytes under MaxTxSize is not projected as fitting
+// when the signed form would just exceed it. 8 bytes is conservative — enough to
+// cover a multi-byte array-length header for large witness counts.
+const witnessSetOverhead = 8
 
 // completeForcingPrefix escapes the min-UTxO dust-change dead-zone by forcing a
 // growing, largest-first prefix of the wallet's UTxOs as explicit inputs,
@@ -550,7 +558,7 @@ func (s *Service) completeForcingPrefix(
 		// Converged on the minimal forced set. Reject it locally if signing it
 		// would exceed the chain's MaxTxSize, turning a would-be submit-time
 		// failure into a clear, pre-approval error.
-		if err := s.guardTxSize(a, utxoAddr); err != nil {
+		if err := s.guardTxSize(ctx, a, utxoAddr); err != nil {
 			return nil, err
 		}
 		return a, nil
@@ -570,8 +578,8 @@ func (s *Service) completeForcingPrefix(
 // in Confirm), not once per input, so the projection counts distinct signing
 // addresses via utxoAddr — inputs sharing an address share one witness. An input
 // whose address is unknown is conservatively assumed to need its own witness.
-func (s *Service) guardTxSize(a *apollo.Apollo, utxoAddr map[string]string) error {
-	pp, err := s.chain.ProtocolParams()
+func (s *Service) guardTxSize(ctx context.Context, a *apollo.Apollo, utxoAddr map[string]string) error {
+	pp, err := backend.ProtocolParamsContext(ctx, s.chain)
 	if err != nil {
 		return fmt.Errorf("protocol params: %w", err)
 	}
@@ -601,7 +609,10 @@ func (s *Service) guardTxSize(a *apollo.Apollo, utxoAddr map[string]string) erro
 	if missingWitnesses < 0 {
 		missingWitnesses = 0
 	}
-	signedSize := len(cborBytes) + missingWitnesses*vkeyWitnessSizeEstimate
+	signedSize := len(cborBytes)
+	if missingWitnesses > 0 {
+		signedSize += missingWitnesses*vkeyWitnessSizeEstimate + witnessSetOverhead
+	}
 	if signedSize > pp.MaxTxSize {
 		return fmt.Errorf(
 			"%w: transaction would exceed the maximum size (%d/%d bytes); reduce the amount or consolidate UTxOs",
--- a/ui/internal/spend/spend.go
+++ b/ui/internal/spend/spend.go
@@ -407,6 +407,15 @@ func (s *Service) Build(ctx context.Context, req SendRequest) (Preview, error) {
 		}
 	}
 
+	// Reject locally if signing this would exceed the chain's MaxTxSize, turning a
+	// would-be submit-time rejection into a clear, pre-approval error. This covers
+	// BOTH paths: the forced-prefix retry can consume extra inputs, but an
+	// ordinary coin-selected send of a large amount can select enough inputs to
+	// cross the limit on its own without ever entering the dead-zone.
+	if err := s.guardTxSize(ctx, a, utxoAddr); err != nil {
+		return Preview{}, err
+	}
+
 	// Store the pending entry (sweeping any that have outlived their TTL first).
 	id := s.mkID()
 	s.mu.Lock()
@@ -555,12 +564,9 @@ func (s *Service) completeForcingPrefix(
 			}
 			return nil, mapCompleteErr(err)
 		}
-		// Converged on the minimal forced set. Reject it locally if signing it
-		// would exceed the chain's MaxTxSize, turning a would-be submit-time
-		// failure into a clear, pre-approval error.
-		if err := s.guardTxSize(ctx, a, utxoAddr); err != nil {
-			return nil, err
-		}
+		// Converged on the minimal forced set. The MaxTxSize check lives in Build
+		// so it covers the coin-selected path too; growing the prefix further
+		// would only add inputs, so there is nothing to gain by checking here.
 		return a, nil
 	}
 
@@ -611,7 +617,15 @@ func (s *Service) guardTxSize(ctx context.Context, a *apollo.Apollo, utxoAddr ma
 	}
 	signedSize := len(cborBytes)
 	if missingWitnesses > 0 {
-		signedSize += missingWitnesses*vkeyWitnessSizeEstimate + witnessSetOverhead
+		signedSize += missingWitnesses * vkeyWitnessSizeEstimate
+		// The one-time witness-set framing is only NEW when the completed tx has
+		// no vkey witnesses yet. If it already carries some, that framing is
+		// already inside cborBytes and adding it again would double-count.
+		// Complete() always leaves the set empty today, so this is a guard
+		// against the projection drifting if that ever changes.
+		if existingWitnesses == 0 {
+			signedSize += witnessSetOverhead
+		}
 	}
 	if signedSize > pp.MaxTxSize {
 		return fmt.Errorf(
--- a/ui/internal/spend/spend.go
+++ b/ui/internal/spend/spend.go
@@ -382,46 +382,29 @@ func (s *Service) Build(ctx context.Context, req SendRequest) (Preview, error) {
 		return Preview{}, fmt.Errorf("%w: lovelace: %w", ErrInvalidRequest, err)
 	}
 	// Complete with the selected input payment key hashes embedded as Conway
-	// required signers. The first pass discovers selected inputs; later passes
-	// rebuild the tx with that signer set so fee estimation covers the bound body.
-	var a *apollo.Apollo
-	var required []lcommon.Blake2b224
-	maxAttempts := len(acct.ReceiveAddresses) + 2
-	for attempt := 0; attempt < maxAttempts; attempt++ {
-		next := apollo.New(s.chain).
-			SetWallet(apollo.NewExternalWallet(changeAddr)).
-			SetChangeAddress(changeAddr).
-			SetFeePadding(feePaddingLovelace).
-			AddLoadedUTxOs(loaded...)
-		for _, kh := range required {
-			next = next.AddRequiredSigner(kh)
+	// required signers (the first pass discovers selected inputs; later passes
+	// rebuild the tx with that signer set so fee estimation covers the bound body).
+	//
+	// A send that would leave change just below the min-UTxO floor puts Apollo in
+	// the "dust-change dead-zone": adding a change output raises the fee estimate,
+	// which drops the change below the floor, which removes the output, which
+	// lowers the fee — Apollo's evaluation loop oscillates and Complete() reports
+	// "evaluation transaction did not converge". When that happens, retry forcing
+	// a growing, largest-first prefix of the loaded UTxOs in as explicit inputs
+	// until the combined change clears the floor (see completeForcingPrefix). This
+	// pulls in only the minimum extra inputs needed — usually one, matching the
+	// legacy builder's "silently pull in another input" behaviour — rather than
+	// forcing the whole wallet, which would blow past MaxTxSize on large wallets.
+	addrCount := len(acct.ReceiveAddresses)
+	a, err := s.completeSend(ctx, changeAddr, recvAddr, lovelace, units, loaded, utxoAddr, addrCount, false)
+	if err != nil {
+		if !isNonConvergenceError(err) {
+			return Preview{}, mapCompleteErr(err)
 		}
-		next = next.PayToAddress(recvAddr, int64(lovelace), units...) //nolint:gosec // validated by parseAmount
-
-		next, err = next.WithContext(ctx).Complete()
-		if err != nil {
-			if isInsufficientFundsError(err) {
-				return Preview{}, fmt.Errorf("%w: %w", ErrInsufficientFunds, err)
-			}
-			return Preview{}, fmt.Errorf("complete transaction: %w", err)
-		}
-
-		tx := next.GetTx()
-		if tx == nil {
-			return Preview{}, errors.New("completed tx is nil")
-		}
-		actual, err := requiredPaymentKeyHashesForInputs(tx.Body.Inputs(), utxoAddr)
+		a, err = s.completeForcingPrefix(ctx, changeAddr, recvAddr, lovelace, units, loaded, utxoAddr, addrCount)
 		if err != nil {
 			return Preview{}, err
 		}
-		if sameKeyHashSet(required, actual) {
-			a = next
-			break
-		}
-		required = actual
-	}
-	if a == nil {
-		return Preview{}, errors.New("required signer set did not converge")
 	}
 
 	// Store the pending entry (sweeping any that have outlived their TTL first).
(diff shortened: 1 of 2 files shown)


ui/internal/spend/spend.go
```diff
@@ -445,6 +439,71 @@ func (s *Service) Build(ctx context.Context, req SendRequest) (Preview, error) {
...
+	var required []lcommon.Blake2b224
+	maxAttempts := len(loaded) + 2
+	for attempt := 0; attempt < maxAttempts; attempt++ {
+		next := apollo.New(s.chain).
+			SetWallet(apollo.NewExternalWallet(changeAddr)).
+			SetChangeAddress(changeAddr).
+			SetFeePadding(feePaddingLovelace)
+		if forceInputs {
+			for _, u := range loaded {
+				next = next.AddInput(u)
```
chrisguiney, line 481: **This is the blocking one: the retry consumes the entire wallet, which crosses `MaxTxSize`.**

`loaded` is every UTxO from every receive address (built at `spend.go:351-369`), so `forceInputs=true` puts the whole wallet into one transaction with no coin selection. Measured at the production `addressWindow = 20` against the repo's own `MaxTxSize: 16384` fixture:

| Wallet UTxOs | Inputs | Fee (4 ADA send) | Signed tx | vs 16384 |
|---|---|---|---|---|
| 19 | 19 | 0.308 ADA | 3,334 B | 20% |
| 380 | 380 | 0.879 ADA | 16,332 B | 100% |
| 418 | 418 | 0.940 ADA | 17,700 B | **108%** |
| 570 | 570 | 1.185 ADA | 23,286 B | **142%** |

Past ~400 UTxOs `Build` succeeds, `Confirm` signs, and the oversized tx reaches `SubmitTx` — the fake chain doesn't validate size, so tests pass while a real node would reject it, after the user has already approved.

The comment just above describes the legacy behavior as pulling in *"another input"*, singular, and that's all this needs: with two 5-ADA inputs against a 4-ADA send, change is ~5.8 ADA — comfortably clear of the floor. Consuming 418 inputs to fix a ~0.14 ADA shortfall is far more than the situation calls for, and it also links every address in the wallet in a single transaction.

A minimal version of the same idea would be to add loaded UTxOs one at a time (largest-first, say) and re-run `Complete()` until it converges, rather than forcing all of them in one shot. That keeps the fix while bounding fee, size, and address linkage. Guarding on `MaxTxSize` before returning would also turn the remaining over-limit case into a clear error rather than a submit-time failure.

Wallets reach several hundred UTxOs through ordinary use (faucets, airdrops, many small receipts), so this is reachable rather than theoretical.

ui/internal/spend/spend_test.go
```diff
@@ -283,6 +284,80 @@ func TestBuildInsufficientFunds(t *testing.T) {
...
+
+	ctx := context.Background()
+	pv, err := s.Build(ctx, SendRequest{To: recvAddr, Lovelace: "4000000"})
+	if err != nil {
+		t.Fatalf("Build in dust-change dead-zone should now succeed, got: %v", err)
+	}
+	// The retry forces the other UTxO in, so both inputs are consumed and the
+	// combined change clears the min-UTxO floor.
+	if len(pv.Inputs) != 2 {
+		t.Fatalf("expected retry to include both inputs, got %d: %v", len(pv.Inputs), pv.Inputs)
```
chrisguiney, line 314: Small note on why the scaling issue above doesn't show up here: this test's wallet holds exactly two UTxOs, which is the one size where "force every input in" and "add one more input" produce identical results. So the assertion passes under either strategy and can't distinguish them.

A case with ~50 UTxOs asserting that `len(pv.Inputs)` stays *small* (rather than equal to the wallet size) would pin the intended behavior and catch a regression toward whole-wallet consolidation.

The two tests you did add are well chosen otherwise — I especially like that this one carries through to `Confirm` to prove the tx actually signs and submits, rather than stopping at `Build`.

ui/internal/spend/spend.go
```diff
@@ -445,6 +439,71 @@ func (s *Service) Build(ctx context.Context, req SendRequest) (Preview, error) {
...
+	ctx context.Context,
+	changeAddr, recvAddr lcommon.Address,
+	lovelace uint64,
+	units []apollo.Unit,
+	loaded []lcommon.Utxo,
+	utxoAddr map[string]string,
+	forceInputs bool,
+) (*apollo.Apollo, error) {
+	var required []lcommon.Blake2b224
+	maxAttempts := len(loaded) + 2
```
chrisguiney: This bound changed from `len(acct.ReceiveAddresses) + 2` to `len(loaded) + 2`, which isn't purely an extraction — the PR describes this as moving the loop into `completeSend`, so it's worth calling out explicitly (or reverting if it wasn't intended).

Tying it to the loaded UTxOs is arguably the more accurate source, since the required-signer set is derived from the selected inputs rather than from every derived address. The practical effect is the ceiling: previously 22 with the production `addressWindow = 20`, now up to 382 for a 380-UTxO wallet, and each attempt is a full Apollo `Complete()`.

In the `forceInputs=true` path the inputs are fixed, so the signer set stabilizes on the second pass and the ceiling never matters. It's the coin-selection path where the two could differ, since adding signers changes the fee, which can change the selection. Not a defect I can point at — just a larger worst case than before, and easy to miss in review.

ui/internal/spend/spend.go
```diff
@@ -445,6 +439,71 @@ func (s *Service) Build(ctx context.Context, req SendRequest) (Preview, error) {
...
+		required = actual
+	}
+	return nil, errors.New("required signer set did not converge")
+}
+
+// isNonConvergenceError reports whether err is Apollo's fee/evaluation
+// non-convergence error. It surfaces in the min-UTxO dust-change dead-zone,
+// where leaving change just under the min-UTxO floor makes Apollo's fee estimate
+// oscillate across the boundary.
+func isNonConvergenceError(err error) bool {
```
chrisguiney, line 663: Non-blocking, but worth a comment for whoever hits it next: this couples the fix to an Apollo error string, and `ui/go.mod` pins Apollo to a pseudo-version (`v2.0.0-20260729192515-7fa2cfd06025`). If a future bump rewords that message, the retry silently stops firing and users are back to seeing the raw non-convergence error, with no test failing to signal it.

If Apollo exports (or would accept) a sentinel error for this, `errors.Is` would be more durable. Failing that, a short test asserting the exact string still appears — or just a comment naming the Apollo version this was verified against — would make the coupling visible at upgrade time.

One adjacent thing worth being aware of: `completeSend` returns its own `errors.New("required signer set did not converge")`, which also contains "did not converge". It doesn't match today because this check uses the full `"evaluation transaction did not converge"` phrase — but if anyone ever shortens the matcher, bursa's own signer-convergence error would be misread as the dust dead-zone and trigger a whole-wallet retry. Might be worth a note here so that stays deliberate.

ui/internal/spend/spend.go
```diff
@@ -382,46 +382,40 @@ func (s *Service) Build(ctx context.Context, req SendRequest) (Preview, error) {
...
+	// "evaluation transaction did not converge". When that happens, retry once
+	// forcing every loaded UTxO in as an explicit input so the combined change
+	// clears the floor (the legacy builder resolved this by silently pulling in
+	// another input).
+	a, err := s.completeSend(ctx, changeAddr, recvAddr, lovelace, units, loaded, utxoAddr, false)
+	if err != nil {
+		switch {
+		case isInsufficientFundsError(err):
+			return Preview{}, fmt.Errorf("%w: %w", ErrInsufficientFunds, err)
+		case isNonConvergenceError(err):
```
chrisguiney: Minor readability point, entirely optional: the inner switch repeats two of the three arms from the outer one — `isInsufficientFundsError` → `ErrInsufficientFunds`, and `default` → `"complete transaction: %w"`. Only the `isNonConvergenceError` arm actually differs between the first attempt and the retry.

Something like a small `mapCompleteErr(err) error` helper for the shared arms, with the retry handling only its own non-convergence case, would flatten the nesting a bit. Purely cosmetic — happy for this to be ignored.

chrisguiney (changes requested):
Thanks for chasing this one down — the dead-zone diagnosis is genuinely good detective work, and the write-up of the oscillation (change output raises the fee, which drops change below the floor, which removes the output, which lowers the fee) made the root cause easy to follow. The extraction into `completeSend` is a clean refactor, and I like that the unresolvable case now gets an actionable error instead of Apollo's raw iteration message.

I'm requesting changes on one thing: the retry escapes the dead zone by consuming **every** UTxO in the wallet, and I was able to measure that breaking at production parameters.

I ran this against `addressWindow = 20` (the production constant in `spend.go:174`), using the repo's own fixture value `MaxTxSize: 16384`:

| Wallet UTxOs | Inputs consumed | Fee on a 4 ADA send | Signed tx | vs 16384 |
|---|---|---|---|---|
| 19 | 19 | 0.308 ADA | 3,334 B | 20% |
| 380 | 380 | 0.879 ADA | 16,332 B | 100% |
| 418 | 418 | 0.940 ADA | 17,700 B | **108% — over** |
| 570 | 570 | 1.185 ADA | 23,286 B | **142% — over** |

Two consequences:

1. **Past roughly 400 UTxOs the transaction is unsubmittable, and nothing catches it locally.** `Build` succeeds, `Confirm` signs, and the oversized tx goes to `SubmitTx`. The fake chain doesn't validate size so tests stay green, but a real node rejects it — after the user has already reviewed and approved.
2. **The fee scales with wallet size rather than send size.** 0.879 ADA on a 4 ADA send is ~22% of the amount, against ~0.18 ADA on the normal path, with nothing in the Preview explaining why.

Worth noting `inputs == walletUTxOs` in every row: the retry fires regardless of how many UTxOs are available, so this is the common path whenever the amount lands in the dead zone — not a rare-wallet edge case.

Verification on my side: `go test -race ./internal/spend/...` passes, and all 15 CI checks pass. Details and a suggested direction are in the inline comments.

wolf31o2:
Thanks for the measurements — the whole-wallet consolidation was the right thing to block. Pushed `e583497` (rebased onto current `main`).

**Blocking: retry no longer consumes the wallet.** The dead-zone retry now forces UTxOs **largest-first, one at a time, stopping at the smallest prefix that converges** (`completeForcingPrefix`) instead of forcing every loaded UTxO in one shot. This bounds inputs, fee, tx size, and address linkage to the minimum the shortfall actually needs — usually the single "another input" the legacy builder pulled in.

Added a **`MaxTxSize` guard** (`guardTxSize`) before returning: it reads the chain's `MaxTxSize`, serializes the completed tx, and projects the signed size (Apollo leaves the witness set empty, so +102 B per input for the vkey witness). Over the limit now returns a clear pre-approval error ("…would exceed the maximum size (N/limit bytes); reduce the amount or consolidate UTxOs") rather than reaching `SubmitTx`.

Re-running your methodology (`addressWindow = 20`, `MaxTxSize: 16384`, 4-ADA dead-zone send):

| Wallet UTxOs | Inputs consumed | Signed tx | vs 16384 |
|---|---|---|---|
| 19  | 2 | 462 B | 2.8% |
| 380 | 2 | 462 B | 2.8% |
| 418 | 2 | 462 B | 2.8% |
| 570 | 2 | 462 B | 2.8% |

Inputs/fee/size are now flat regardless of wallet size, and no oversized tx is produced.

**Non-blocking points:**

- **2-UTxO test couldn't distinguish the strategies** — added `TestBuildDeadZoneForcesBoundedInputs`: 50-UTxO wallet, dead-zone amount, asserts `len(pv.Inputs)` stays small (≤5) and `!= walletUTxOs`, the tx is under `MaxTxSize`, and it carries through `Confirm` to prove it still signs and submits.
- **`maxAttempts` bound** — kept `len(loaded) + 2`, now with a comment on what bounds the signer-convergence loop (distinct signers ≤ inputs ≤ `len(loaded)`, plus the confirming pass) for both selection and forced modes.
- **Apollo error-string coupling** — `isNonConvergenceError`'s doc now names the pinned Apollo version (`v2.0.0-20260729192515-7fa2cfd06025`) it was verified against, and notes that `completeSend`'s own "required signer set did not converge" also contains "did not converge", so the matcher must keep the full "evaluation transaction did not converge" phrase. If Apollo ever exports a sentinel for this, happy to switch to `errors.Is`.
- **Nested switch** — flattened the shared insufficient-funds/default arms into a `mapCompleteErr` helper used by both `Build` and the retry.

`go vet`, `go test -race ./internal/spend/...`, and `golangci-lint` are clean; `ui-go` and `ui-lint` pass in CI. No rendered UI change (coin-selection logic only), so nothing to screenshot.

wolf31o2:
Addressed the CodeRabbit nitpicks in the same commit:

- **`guardTxSize` witness count** — kept the per-input estimate (it's the conservative direction for a size guard — over-estimating only rejects a shade earlier, never lets an oversized tx through) and rewrote the doc comment to describe it as a deliberate upper bound rather than "one fake witness per signer", since `Confirm` signs once per distinct input address.
- **`maxAttempts` bound** — tightened to `min(len(loaded), receiveAddressCount) + 2`. Distinct signers can't exceed the number of derived receive addresses either, so a 400-UTxO / 20-address wallet now caps at ~22 `Complete()` calls instead of ~402. This also restores the address-count relationship @chrisguiney flagged.
- **Rejection-branch coverage** — added `TestBuildDeadZoneRejectsOversizeTx` (sets `MaxTxSize` below the projected size, asserts `Build` returns `ErrInsufficientFunds` containing "exceed the maximum size" and never calls `SubmitTx`) and `TestBuildDeadZoneSizeGuardDisabled` (pins the `MaxTxSize: 0` disabled-guard path).

`go vet`, `go test -race ./internal/spend/...`, and `golangci-lint` are clean.

ui/internal/spend/spend.go
```diff
@@ -382,67 +382,249 @@ func (s *Service) Build(ctx context.Context, req SendRequest) (Preview, error) {
...
+	if err != nil {
+		return fmt.Errorf("encode tx for size check: %w", err)
+	}
+	tx := a.GetTx()
+	numInputs, existingWitnesses := 0, 0
+	if tx != nil {
+		numInputs = len(tx.Body.Inputs())
+		existingWitnesses = len(tx.WitnessSet.VkeyWitnesses.Items())
+	}
+	missingWitnesses := numInputs - existingWitnesses
```
chrisguiney: **This is the blocking one, and it's the last thing between this and an approval.**

This counts one witness per *input*, but `Confirm` signs once per *distinct address* (step 5/6, spend.go:1146-1161), so the signed tx carries one witness per distinct address. On a single-address wallet — the common shape for exactly the dust-heavy wallets that end up here — that inflates the projection by `(inputs - 1) * 102` bytes.

I measured it by disabling the guard and comparing the projection against the actual CBOR that `Confirm` hands to `SubmitTx`. Wallet of 300 x 1.5 ADA UTxOs, all on `addr0`, with the repo's own `MaxTxSize: 16384`:

| Send | Inputs | Projected | Actual signed | |
|---|---|---|---|---|
| 170 ADA | 115 | 16,057 | 4,433 | ok |
| 200 ADA | 135 | 18,817 | **5,153** | rejected |
| 300 ADA | 201 | 27,925 | **7,529** | rejected |
| 600 ADA | 402 | 55,664 | **14,766** | rejected |

The guard cuts the wallet off around 120 inputs when the real ceiling sits near 450, and the user is told the transaction "would exceed the maximum size" when it is under a third of the limit — then advised to consolidate UTxOs, which is the thing they were already trying to do.

The part I'd push back on is the comment just above, at lines 566-568: *"Over-estimating only ever makes the guard reject a shade earlier."* At 3.7x it isn't a shade, and that framing is what makes the over-count read as harmless. To be clear about where this came from — CodeRabbit reviewed this exact spot and said over-counting was the safe direction and no behavior change was needed. That advice was wrong, so this isn't on you.

Reusing the lookup `Confirm` already does fixes it. I tried this locally and the 200 ADA case then builds and submits at 135 inputs with a 0.388 ADA fee:

```suggestion
	tx := a.GetTx()
	numSigners, existingWitnesses := 0, 0
	if tx != nil {
		// Confirm signs once per DISTINCT input address, so the signed tx carries
		// one vkey witness per distinct address, not one per input.
		seen := make(map[string]bool)
		for _, inp := range tx.Body.Inputs() {
			ref := hex.EncodeToString(inp.Id().Bytes()) + "#" + strconv.Itoa(int(inp.Index()))
			if addr, ok := utxoAddr[ref]; ok {
				seen[addr] = true
			} else {
				numSigners++ // unknown input: assume its own witness
			}
		}
		numSigners += len(seen)
		existingWitnesses = len(tx.WitnessSet.VkeyWitnesses.Items())
	}
	missingWitnesses := numSigners - existingWitnesses
```

That needs `utxoAddr map[string]string` threaded into `guardTxSize` — the caller at line 548 already has it in scope. Take it in whatever shape reads best to you; the shape matters less than counting addresses rather than inputs.

One adjacent thing, explicitly **not** blocking: the guard only covers the forced-prefix path, not ordinary coin selection. That asymmetry predates this PR and I'd leave it alone here — just naming it so it stays a deliberate choice rather than an oversight.

ui/internal/spend/spend.go
```diff
@@ -382,67 +382,249 @@ func (s *Service) Build(ctx context.Context, req SendRequest) (Preview, error) {
...
+	lovelace uint64,
+	units []apollo.Unit,
+	loaded []lcommon.Utxo,
+	utxoAddr map[string]string,
+	addrCount int,
+) (*apollo.Apollo, error) {
+	sorted := make([]lcommon.Utxo, len(loaded))
+	copy(sorted, loaded)
+	sort.SliceStable(sorted, func(i, j int) bool {
+		return sorted[i].Output.Amount().Uint64() > sorted[j].Output.Amount().Uint64()
```
chrisguiney, line 553: Non-blocking observation for a follow-up, not this PR. `Output.Amount()` is lovelace only (`Assets()` is separate), so the ordering here is largest-lovelace-first. For a plain ADA send that's exactly right and the measurements bear it out.

For a token send it means the prefix grows purely by lovelace until it happens to reach the UTxO holding the asset — if that one is small, the prefix can walk most of the wallet to get there. The `asset underflow` arm of `isInsufficientFundsError` keeps it correct, and `guardTxSize` keeps it from producing something unsubmittable, so this is bounded rather than broken. Might be worth a short comment noting that asset sends order by lovelace, so the next person doesn't have to rediscover it.

ui/internal/spend/spend_test.go
```diff
@@ -283,6 +285,211 @@ func TestBuildInsufficientFunds(t *testing.T) {
...
+// signedTxSizeEstimate mirrors Service.guardTxSize: completed CBOR plus one vkey
+// witness per input (Complete leaves the witness set empty).
+func signedTxSizeEstimate(t *testing.T, a *apollo.Apollo) int {
+	t.Helper()
+	cborBytes, err := a.GetTxCbor()
+	if err != nil {
+		t.Fatalf("GetTxCbor: %v", err)
+	}
+	return len(cborBytes) + len(a.GetTx().Body.Inputs())*vkeyWitnessSizeEstimate
+}
```
chrisguiney: Non-blocking, but this is why the issue above survived the test suite: this helper mirrors `guardTxSize`'s arithmetic rather than checking it, so it reproduces the same per-input over-count. A test that re-derives the implementation's formula can confirm the formula was applied consistently, but it can't tell you the formula is wrong.

`fakeChain.SubmitTx` already records the real signed bytes in `fc.submitCbor`, so asserting against that after `Confirm` would measure the thing that actually matters — and would have flagged the 3.7x gap on its own.

ui/internal/spend/spend_test.go
```diff
@@ -283,6 +285,211 @@ func TestBuildInsufficientFunds(t *testing.T) {
...
+// bounded input set well under the wallet size, produce a tx under MaxTxSize,
+// and still sign + submit.
+func TestBuildDeadZoneForcesBoundedInputs(t *testing.T) {
+	acct := mustDeriveConfirmAccount(t)
+	addr0 := acct.ReceiveAddresses[0]
+	recvAddr := acct.ReceiveAddresses[2]
+
+	const walletUTxOs = 50
+	fc := newFakeChain(5_000_000, addr0)
+	fillDeadZoneUTxOs(fc, walletUTxOs-1, addr0) // newFakeChain already seeded one
```
chrisguiney, line 385: Non-blocking. This is a good test and it does pin the bound I asked for — thank you. One gap worth knowing about: every UTxO here is 5 ADA (both `newFakeChain` and `fillDeadZoneUTxOs` use `5_000_000`), so `sort.SliceStable` is a no-op on this input. I deleted the sort in `completeForcingPrefix` entirely and the whole package still passes.

So the *bounded* part is covered, but the *largest-first* part — which is what keeps the prefix at one extra input in the common case — isn't. A mixed-size case pins it cheaply: thirty 5-ADA UTxOs plus a single 100-ADA one loaded last should converge at exactly 1 input. I ran that against this branch and it does, so it would go green as-is and fail if the ordering ever regressed.

ui/internal/spend/spend_test.go
```diff
@@ -283,6 +285,211 @@ func TestBuildInsufficientFunds(t *testing.T) {
...
+// dead-zone retry must fail Build with an ErrInsufficientFunds carrying the
+// size-limit message — never return a Preview.
+func TestBuildDeadZoneRejectsOversizeTx(t *testing.T) {
+	acct := mustDeriveConfirmAccount(t)
+	addr0 := acct.ReceiveAddresses[0]
+	recvAddr := acct.ReceiveAddresses[2]
+
+	fc := newFakeChain(5_000_000, addr0)
+	fillDeadZoneUTxOs(fc, 9, addr0) // 10 UTxOs total on addr0
+	fc.pp.MaxTxSize = 400           // below any completed body + witness projection
```
chrisguiney: Non-blocking, but worth flagging now because it interacts with the fix above: `MaxTxSize = 400` is tight enough that this test passes *because of* the per-input over-count. Once witnesses are counted per distinct address, the projection for this 2-input single-address tx drops below 400 and `Build` succeeds, so the test fails.

That's not a problem with the test's intent — the branch is worth covering, and I'm glad it's here. It just needs recalibrating alongside the change. Sizing the threshold off the measured body (something like `len(cbor) + 1*vkeyWitnessSizeEstimate - 1`) rather than a hardcoded 400 would keep it pinned to the branch rather than to the counting method.

chrisguiney (changes requested):
Thanks for the rework — this is a big improvement, and the core of my last round is genuinely fixed. I re-ran the exact wallet sizes I measured before, against this commit:

| Wallet UTxOs | Inputs (was) | Fee on a 4 ADA send (was) | Signed tx (was) |
|---|---|---|---|
| 19 | **2** (19) | **0.177 ADA** (0.308) | **462 B** (3,334) |
| 380 | **2** (380) | **0.177 ADA** (0.879) | **462 B** (16,332) |
| 418 | **2** (418) | **0.177 ADA** (0.940) | **462 B** (17,700) |
| 570 | **2** (570) | **0.177 ADA** (1.185) | **462 B** (23,286) |

The fee is now flat at ~0.177 ADA — in line with the normal path's ~0.18 — instead of scaling with wallet size, and the input set stays at 2 no matter how large the wallet gets. I also checked that largest-first does real work: with one 100-ADA UTxO among thirty 5-ADA ones, it converges at k=1 and consumes a single input.

The other four points are addressed too. The `maxAttempts` bound is sound — `loaded` is built only from `acct.ReceiveAddresses` (spend.go:352), so `addrCount` is a valid ceiling on distinct signers. I confirmed the exact Apollo phrase is present at `apollo.go:1748` / `1755` in the pinned version, and the note about never shortening the matcher is a good catch to leave behind for whoever bumps it next. `mapCompleteErr` reads well.

I'm requesting changes on one new thing: the size guard counts one witness per *input*, but `Confirm` signs once per *distinct address*, so it over-projects by ~3.7x and rejects transactions that are comfortably under the limit. Details, measurements, and a fix I verified are inline on `guardTxSize`.

Worth saying up front: that one isn't a regression — those sends already failed on `main` — and CodeRabbit explicitly told you over-counting was the safe direction and needed no behavior change, which turned out not to hold. The rest of the inline comments are non-blocking.

Verification on my side: `go test -race ./internal/spend/...` passes locally, all 18 CI checks pass, and I found no concurrency concerns — the sort operates on a copy, there's no new shared state, and `s.mu` usage is unchanged. No performance concern either: insufficient-funds prefixes short-circuit after a single `Complete()`, so even a 570-UTxO wallet builds in under 2 ms.

wolf31o2:
Fixed the witness over-count in `3162c66`.

**Blocking — `guardTxSize` now counts distinct signing addresses, not inputs.** `Confirm` signs once per distinct input address, so the projection now walks the inputs, maps each to its address via `utxoAddr` (an unknown input conservatively assumed to need its own witness), and counts distinct addresses. On a single-address dust wallet that's one witness instead of one-per-input, so the ~3.7× inflation is gone and sends that fit are no longer rejected. I also rewrote the comment that framed the over-count as "a shade earlier" — thanks for the measurements and the verified suggestion.

Non-blocking points, all addressed:
- **Test measures the real thing** — `TestBuildDeadZoneForcesBoundedInputs` now asserts against the actual signed bytes `SubmitTx` receives (`submittedTxCbor()`) instead of a helper that re-derived the guard's own formula; deleted that helper. This is what would have caught the 3.7× gap.
- **Largest-first is now pinned** — added `TestBuildDeadZoneForcesLargestFirst`: thirty 5-ADA UTxOs plus one 100-ADA loaded last must converge at exactly 1 input. Deleting the sort makes it fail (the all-5-ADA case couldn't).
- **Oversize threshold pinned to the branch** — `TestBuildDeadZoneRejectsOversizeTx` now measures the completed body with the guard disabled and sets `MaxTxSize` to `body + 1 witness − 1`, so it stays valid regardless of the counting method rather than depending on the old over-count.
- **Lovelace-ordering note** — added a comment that `Output.Amount()` is lovelace-only, so a token send grows the prefix by lovelace until it reaches the asset-bearing UTxO (bounded by the asset-underflow arm and `guardTxSize`).
- The guard-only-covers-forced-prefix asymmetry: left as-is per your note, deliberately.

`go vet`, `go test ./internal/spend/...`, and `golangci-lint` are clean.

wolf31o2:
Triaged the cubic findings — two fixed in `89d11f6`, two deferred to match the human review's explicit scoping.

**Fixed**
- **#4 (context-aware params)** — `guardTxSize` now looks up protocol params via `backend.ProtocolParamsContext(ctx, s.chain)` (the same helper `multisig` uses) instead of the context-less `s.chain.ProtocolParams()`, so a canceled `Build` is honored during the forced retry. Threaded `ctx` in.
- **#2 (witness-set overhead undercount)** — the projection added `N × vkeyWitnessSizeEstimate` but not the one-time CBOR framing that appears when the empty witness set gains its first vkey-witness field (map key + array header). Added a conservative one-time `witnessSetOverhead` (8 B) so a tx sitting a few bytes under `MaxTxSize` isn't projected as fitting when the signed form would just exceed it. Kept it one-time (not per-witness) so it doesn't reintroduce over-rejection.

**Deferred (matching the human reviewer's explicit decisions on this PR)**
- **#1 (guard the normal Build path too)** — the human review explicitly chose to leave this asymmetry as a deliberate choice for this PR ("the guard only covers the forced-prefix path… that asymmetry predates this PR and I'd leave it alone here"). Keeping it scoped as agreed rather than expanding it here.
- **#3 (asset-bearing UTxO ordering, P1)** — the human review assessed this as "bounded rather than broken" (the asset-underflow arm keeps it correct and `guardTxSize` prevents anything unsubmittable) and explicitly scoped it as "a follow-up, not this PR." A comment noting the lovelace-ordering behaviour for token sends is already in `completeForcingPrefix`. A proper asset-first selection (with token-send tests) is worth its own change.

`go build`, `go vet`, `go test ./internal/spend/...`, and `golangci-lint` are clean.

wolf31o2:
Worked through the outstanding review points in `f245f70`.

**Fixed — size guard missed the coin-selected path.** `guardTxSize` only ran inside `completeForcingPrefix`, so it covered the dust-change retry but not an ordinary send. Confirmed reachable before fixing: a 200 ADA send from a 40 x 10 ADA wallet leaves change far clear of the min-UTxO floor, so it never enters the dead-zone — and against the previous code `Build` returned `nil` on a transaction over `MaxTxSize`, which would then have failed at `SubmitTx` after approval. The check now lives in `Build` and covers both paths; the redundant call in `completeForcingPrefix` is gone. Pinned by `TestBuildRejectsOversizeTxOnPlainPath`, which fails against the previous commit.

**Fixed — witness-set overhead double-count.** `witnessSetOverhead` is now only added when the completed tx has no vkey witnesses yet. Latent rather than live (`Complete()` always leaves the set empty today), but it keeps the projection consistent with what the comment claims.

**Not changed — native-asset input ordering.** Flagged as P1, but the prefix loop treats an asset shortfall as "grow and retry" via the asset-underflow arm of `isInsufficientFundsError`, so a wallet that genuinely holds the asset keeps growing until it reaches it rather than reporting exhaustion. The bounded-but-larger-prefix behaviour for token sends is documented at the sort. Matches the human review on the same lines, which called it a follow-up rather than a blocker.

**Already addressed.** The undercounted CBOR overhead was handled in `89d11f6`. The largest-first ordering test exists as `TestBuildDeadZoneForcesLargestFirst` — verified it genuinely pins the behaviour by deleting the sort, which makes it fail with 2 inputs instead of 1.

Verification: full `ui` suite, `go vet`, and `gofmt` clean. `nilaway` reports 10 findings in this package both with and without the change — all pre-existing, in unrelated functions, and outside the root-module scope CI runs it against.

chrisguiney (approved):

PR #665, opened 2026-08-04
feat(ui): read-only governance action browser

wolf31o2:
Refs #723, #724, and #725

Read-only browser for on-chain Conway governance actions (proposals): type, lifecycle status, and Yes/No/Abstain vote tallies.

## Endpoint

`GET /wallet/governance-actions` — gated like other node reads, no external-consent gate (all data is node-local). Query params `q` (case-insensitive substring over action ID / tx hash / type), `page`, `count`. Response: `{ actions: [{ action_id, tx_hash, action_index, type, proposed_epoch, expires_epoch, status, anchor_url, deposit, yes_votes, no_votes, abstain_votes }], total, page, count }`.

## Data source

Dingo v0.66.2 exposes no Blockfrost-compatible route for governance proposals, so the chain client reads Dingo's local `metadata.sqlite` directly (read-only), the same node-local pattern `AccountDRepID` uses. Reads the `governance_proposal` table (filtered `deleted_slot IS NULL`, newest first) and aggregates the `governance_vote` table into per-proposal tallies. `action_type` maps to type labels (`info` / `treasury-withdrawal` / `param-change` / `hard-fork` / `no-confidence` / `new-committee` / `update-constitution`); status derives from the `enacted`/`ratified`/`expired` epoch columns (else `active`). `action_id` is the CIP-129 bech32 `gov_action1…` id. An absent data dir / metadata file / governance table yields an empty list, never an error.

## Files

Backend:
- `ui/internal/chain/blockfrost.go` — `GovernanceActions(ctx)` + `GovernanceAction` type, sqlite read + vote-tally aggregation, type/status/id helpers.
- `ui/internal/api/api.go` — `GET /wallet/governance-actions` handler, `NodeLookup.GovernanceActions`, `filterAndPageGovernanceActions` (clamp-both-sides pagination).

Frontend:
- `ui/web/src/screens/Governance.tsx` — searchable/paginated table (action id + copy + explorer link, type, status, proposed epoch, tallies) with loading/empty/error states.
- `ui/web/src/api/types.ts`, `ui/web/src/api/client.ts` (`getGovernanceActions`), `ui/web/src/api/hooks.ts` (`useGovernanceActions`), `ui/web/src/explorer.ts` (`govaction` → `govAction` path), `ui/web/src/app.tsx` (Governance nav route, gated on a queryable node).

## Tests

- `blockfrost_test.go`: `TestGovernanceActionsFromDingoMetadata` (temp `metadata.sqlite`: type/status mapping, tallies, soft-deleted excluded, CIP-129 id), `TestGovernanceActionsNoDataDir`, `TestGovernanceActionsMissingTable`.
- `api_test.go`: `TestGetGovernanceActions` (populated), `…SearchFilter`, `…Empty`, `…PageOverflow` (no-panic), `…NilLookup` — all via `localReq`.
- `Governance.test.tsx`: lists actions, server-side search, empty state, API error.

## Verify

`ui/`: `go build ./...`, `go vet ./...`, `golangci-lint run ./internal/chain/... ./internal/api/...` (0 issues), `go test ./...` pass. `ui/web`: `npm run lint` (0 errors), `type-check`, `npm test` (633 pass), `npm run build`. Repo root: `CGO_ENABLED=0 go build ./...`.

## Screenshots

Governance actions — list (all types / statuses / tallies):

Governance actions list

Filtered (server-side search `q=treasury`):

Governance actions filtered

## Summary by CodeRabbit

- **New Features**
  - Added a read-only Governance screen for browsing Conway governance actions.
  - View action IDs, types, statuses, proposal epochs, vote tallies, deposits, and explorer links.
  - Added server-side search and pagination, with loading, empty, and error states.
  - Added navigation support and full action-ID copy controls.
  - Governance actions are available when supported by the connected node.
  - Added consistent shortening for long identifiers while preserving copyable full IDs.


wolf31o2:
Addressed the bot findings.

**Fixed**
- **Status priority** (`blockfrost.go`): `govActionStatus` checked `ratified` before `expired`, so an action with both epoch columns set was mislabeled "ratified" instead of the terminal "expired". Reordered to `enacted` > `expired` > `ratified` > `active`; added `TestGovActionStatus` covering all four states plus the two priority collisions (there was no prior test for this function).
- **Stale error on retry** (`hooks.ts`): `useGovernanceActions` now clears `error` when starting a new debounced fetch, so a stale banner doesn't linger under the spinner.
- **Dedup** (`format.ts`): moved the duplicated `shortId` into shared `format.ts` (alongside `formatAda` et al.) and updated both `Governance.tsx` and `PoolDirectory.tsx`; added direct unit tests.

No rendered screen changed — the status fix alters a computed string only in an edge-case data combination, and the `shortId` move is identical output. Full `go test`, `vitest` (639), `tsc`, `eslint` clean.

ui/web/src/api/hooks.ts
```diff
@@ -213,6 +215,56 @@ export function useNftMedia(): NftMediaState {
...
+// query or page changes. Node-local read; no polling, since governance actions
+// change on the order of epochs, and the screen offers a manual refresh via
+// navigation.
+export function useGovernanceActions(params: { q: string; page: number }): GovernanceActionsState {
+  const { q, page } = params;
+  const [data, setData] = useState<GovernanceActionsResponse | null>(null);
+  const [error, setError] = useState<Error | null>(null);
+  const [loading, setLoading] = useState(true);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
wolf31o2: Agreed that the duplication is real — the governance hook and `PoolDirectory` run the same debounce/cancel/loading lifecycle independently, and they have already drifted: the governance hook now clears stale data on error (634c2ea) and the DRep screen gained an `aria-busy` stale marker (ae51e6d on #661), neither of which `PoolDirectory` has.

Not folding the extraction into this PR: it would touch three browse screens across two open PRs (#661 and this one) that are still in review, so it belongs after both land rather than as a cross-cutting refactor inside a feature branch. Flagging it here as the follow-up rather than resolving it as done.

ui/internal/chain/blockfrost.go
```diff
@@ -643,6 +644,237 @@ func (c *Client) DRep(ctx context.Context, drepID string) (DRepInfo, error) {
...
+	uint8(lcommon.GovActionTypeNewConstitution):    "update-constitution",
+	uint8(lcommon.GovActionTypeInfo):               "info",
+}
+
+// Governance vote choices as stored by Dingo (models.GovernanceVote.Vote).
+const (
+	dingoGovVoteNo      = 0
+	dingoGovVoteYes     = 1
+	dingoGovVoteAbstain = 2
+)
```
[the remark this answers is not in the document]
wolf31o2: Both halves valid, fixed in 634c2ea.

The local `dingoGovVoteNo/Yes/Abstain` constants now alias Dingo's own `models.VoteNo/VoteYes/VoteAbstain` — `github.com/blinklabs-io/dingo` is already a dependency of the `ui` module (v0.69.0), so this costs nothing and turns an upstream renumbering into a compile error here rather than a silently mis-attributed tally.

Unknown values no longer fall through: the `switch` has a `default` that returns `unknown governance vote choice %d for proposal %d`. Dropping them silently would under-report the tally and make the displayed Yes/No/Abstain numbers disagree with the chain, which is worse than surfacing that this build does not understand a vote kind.

ui/web/src/api/hooks.ts
```diff
@@ -213,6 +215,56 @@ export function useNftMedia(): NftMediaState {
...
+      getGovernanceActions({ q: query, page })
+        .then((d) => {
+          if (!cancelled) {
+            setData(d);
+            setError(null);
+          }
+        })
+        .catch((e: Error) => {
+          if (!cancelled) setError(e);
+        })
```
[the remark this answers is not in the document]
wolf31o2: Valid, fixed in 634c2ea. The catch now clears `data` alongside setting the error, so a failed request no longer leaves the previous query or page rendered as the answer to the new one while an error is displayed above it.

ui/internal/api/api.go
```diff
@@ -1820,6 +1846,63 @@ func filterAndPagePools(pools []chain.PoolInfo, q, pageStr, countStr string) poo
...
+// (q) is a case-insensitive substring match against the action ID, tx hash, and
+// type label. page is 1-based; count is clamped to [1, poolDirMaxCount].
+// Invalid/absent page or count fall back to their defaults rather than erroring.
+func filterAndPageGovernanceActions(actions []chain.GovernanceAction, q, pageStr, countStr string) governanceActionsResponse {
+	needle := strings.ToLower(strings.TrimSpace(q))
+	matched := make([]chain.GovernanceAction, 0, len(actions))
+	for _, a := range actions {
+		if needle == "" ||
+			strings.Contains(strings.ToLower(a.ActionID), needle) ||
+			strings.Contains(strings.ToLower(a.TxHash), needle) ||
```
chrisguiney, line 1867: This re-implements `filterAndPage[T]` rather than calling it.

The generic at `:1798` already does every part of this — the lower-cased needle, the count clamp to `[1, poolDirMaxCount]`, the 1-based page fallback, the both-sides start/end clamp, and the nil-slice guard. `filterAndPagePools` (`:1841`) and `filterAndPageDReps` (`:1922`) each use it in about seven lines; this is the only one that does not, and it carries a verbatim copy of the overflow comment as well.

It is easy to see how it happened, and it is not really a mistake: `git log -S "func filterAndPage[T any]"` puts the generic in `5e6b294` (#661, the DRep directory), which landed on main *after* this branch was cut. A merge will not fold an extraction into code that was written before it existed.

Collapsing to the generic drops roughly 28 lines, and — the part I actually care about — leaves the overflow clamp defined in one place instead of two. A clamp that guards against a panic is exactly the kind of logic that should not have a second copy quietly drifting.

The matcher itself stays as-is; it would just become the callback:

```go
pageItems, total, page, count := filterAndPage(actions, q, pageStr, countStr, func(a chain.GovernanceAction, needle string) bool {
    return strings.Contains(strings.ToLower(a.ActionID), needle) ||
        strings.Contains(strings.ToLower(a.TxHash), needle) ||
        strings.Contains(strings.ToLower(a.Type), needle)
})
```

ui/internal/chain/blockfrost.go
```diff
@@ -676,6 +679,254 @@ func (c *Client) DRep(ctx context.Context, drepID string) (DRepInfo, error) {
...
+		SELECT id, tx_hash, action_index, action_type, proposed_epoch,
+		       expires_epoch, enacted_epoch, ratified_epoch, expired_epoch,
+		       anchor_url, deposit
+		FROM governance_proposal
+		WHERE deleted_slot IS NULL
+		ORDER BY proposed_epoch DESC, id DESC`)
+	if err != nil {
+		return nil, fmt.Errorf("query governance proposals: %w", err)
+	}
+	defer rows.Close()
```
chrisguiney, line 770: Scaling note, not blocking: this query has no `LIMIT` and no cache, and neither does the vote aggregation above it.

Every request reads the whole `governance_proposal` table and runs a full `GROUP BY` across `governance_vote`, then Go filters and pages the result. On mainnet that aggregate is roughly every DRep times every proposal, and the whole thing re-runs on mount, on each debounced keystroke, and on each page change. `Pools` has a one-minute TTL cache (`poolCacheTTL`, `:312`); there is no equivalent here.

The part that gives me more pause than the row count: this runs against Dingo's *live* metadata DB, and `sqliteReadOnlyDSN` sets `busy_timeout(30000)`. If Dingo is holding a write lock during sync, a governance request can sit for a long time on a UI path. That helper is shared and predates this PR, so it is not yours to fix here — but this is the first screen where a user is likely to hit it repeatedly while the node is busy.

Either pushing `LIMIT`/`OFFSET` into SQL or adding a short TTL cache in the shape `Pools` already uses would take the edge off. Worth a follow-up issue rather than growing this PR.

ui/internal/chain/blockfrost.go
```diff
@@ -676,6 +679,254 @@ func (c *Client) DRep(ctx context.Context, drepID string) (DRepInfo, error) {
...
+				"skipping governance vote of an unknown kind",
+				"vote", vote,
+				"proposal_id", proposalID,
+				"count", count,
+			)
+			continue
+		}
+		tallies[proposalID] = t
+	}
+	if err := rows.Err(); err != nil {
```
chrisguiney, line 875: The decision here is right, and the comment explaining it is genuinely good — turning one unrecognised vote into a 503 that blanks the screen would be much worse than under-reporting a tally.

One small thing about the mechanics: this sits inside the `GROUP BY` loop, so it logs once per (proposal, unknown kind) pair. If a future Dingo adds a vote kind, that is potentially hundreds of warning lines per request — and requests fire at debounce frequency while someone types in the search box, so it would not be a one-off burst.

Counting the skipped votes and emitting a single warning after the loop would keep exactly the same signal ("this build saw a vote kind it does not know") without the volume.

ui/web/src/api/types.ts
```diff
@@ -263,6 +263,37 @@ export interface PoolDirectoryResponse {
...
+export interface GovernanceAction {
+  action_id: string;
+  tx_hash: string;
+  action_index: number;
+  type: string;
+  proposed_epoch: number;
+  expires_epoch: number;
+  status: string;
+  anchor_url: string;
+  deposit: string;
```
chrisguiney, line 281: `expires_epoch`, `anchor_url`, and `deposit` are computed, serialized, and typed here, but `Governance.tsx` renders none of them — the table shows action id, type, status, proposed epoch, and tallies. Worth either surfacing them or trimming them from the response, so the API surface matches what the screen actually uses.

`anchor_url` is the one I would think about specifically. It is arbitrary text from an on-chain proposal, so it is attacker-controlled, and it is inert today only because nothing renders it. The obvious next iteration of this screen is a link to the proposal metadata, and at that point a `javascript:` or `data:` URL in an `href` becomes a real problem in a wallet.

A short note on the field — or a scheme allowlist applied when it is read — would save whoever adds that link from having to rediscover the analysis.

ui/web/src/explorer.ts
```diff
@@ -25,6 +25,7 @@ const PATHS: Record<ExplorerKind, string> = {
   address: "address",
   pool: "pool",
   drep: "drep",
+  govaction: "govAction",
```
chrisguiney, line 28: Worth double-checking: this is the only camelCase entry in `PATHS` — `tx`, `address`, `pool`, and `drep` are all lowercase.

That may well be correct for cardanoscan, which is not consistent about this across its own routes. I could not verify a third-party URL scheme offline, and `explorer.test.ts` pins whatever is written here rather than what the site serves, so a wrong path would fail silently at the one moment a user clicks through. Just a quick confirmation against a live governance action would settle it.

chrisguiney (changes requested):
Nice feature, and the care shows in the places that matter for a wallet. The "node-local, so deliberately no external-consent gate" reasoning is spelled out at the handler, the degradation path (absent data dir / metadata file / table all read as empty) is deliberate, and the unknown-vote-kind decision in `ed9b2791` — skip and log rather than 503 the whole screen — is the right call with the right justification attached.

I went looking hard for the two things that would matter most here, and neither is present:

- **No SQL injection.** Every query is a static string; the only bound value anywhere is the table name in `sqliteHasTable` (`blockfrost.go:886`), correctly passed as `?`. Nothing from the request reaches SQL — search and paging happen in Go.
- **No XSS via `anchor_url`.** This was my main worry, since it is attacker-controlled on-chain data. It is returned and typed but never rendered: `Governance.tsx` has no `href=` and no `dangerouslySetInnerHTML`, and the one outbound link goes through the existing `ExplorerLink`, which hardcodes `https://`, `encodeURIComponent`s the id (`explorer.ts:41`), sets `rel="noopener noreferrer"`, and routes through the webview external-open bridge instead of navigating the wallet window.

Other things I checked and found sound: resource handling on both queries (`defer db.Close()`, `defer rows.Close()`, `rows.Err()` after each loop, context propagated); `govActionID` guarding `len(txHash)==32 && index<=255` before building the CIP-129 id, with a sane `hex#index` fallback; the frontend hook being leak-free (250 ms debounce, `cancelled` flag, `clearTimeout` cleanup); and the `shortId` move out of `PoolDirectory.tsx` being byte-identical, so pools cannot regress.

`ui` Go build and tests pass, and `Governance.test.tsx`, `format.test.ts` and `explorer.test.ts` all pass locally. One note so it is not mistaken for a finding: my local web run shows 50 failures, all from `ExperimentalWarning: localStorage is not available` in my Node build — `deviceKind.test.ts` entirely, plus the two `app.test.tsx` operator-mode tests that call `window.localStorage.setItem` directly. Nothing governance-related, and CI is green on all 18 checks.

Only one thing I would ask for before merge, on `api.go:1864`, and it is mechanical. The rest are follow-ups. If you would rather land the feature and take that collapse as a follow-up issue, I have no real objection — say so and I will re-review to approve.

ui/internal/chain/blockfrost.go
wolf31o2: Tracked in #723: https://github.com/blinklabs-io/bursa/issues/723

ui/internal/chain/blockfrost.go
wolf31o2: Tracked in #724: https://github.com/blinklabs-io/bursa/issues/724

ui/web/src/api/types.ts
wolf31o2: Tracked in #725: https://github.com/blinklabs-io/bursa/issues/725

ui/web/src/explorer.ts
wolf31o2: Verified against live Cardanoscan detail pages on all three hosts. The route is `/govAction/<gov_action…>` with the camel-case `A`: [mainnet](https://cardanoscan.io/govAction/gov_action1jxne7hynfd7frcczwumd2eggps4kvy0msjztz9t0mutpy870ksgqqp6vp3p), [preprod](https://preprod.cardanoscan.io/govAction/gov_action1zk80dvjfklp7cgvuvtg37zuwwe4r2erj6q3m67c7wdh0akth70rqq44samn), and [preview](https://preview.cardanoscan.io/govAction/gov_action1cj7e056x39vxg8cnqc9v6drlg3dt3jn92lp2xeqkawu8w0qqvq2qqjtun9u). The existing mapping is correct.

ui/internal/api/api.go
wolf31o2: Addressed in 3e8e810. `filterAndPageGovernanceActions` now delegates the shared paging, overflow-clamping, and non-nil empty-slice behavior to `filterAndPage`; the governance-specific matcher is unchanged.

wolf31o2:
Review update:

- `ui/internal/api/api.go`: governance action filtering and paging now reuse the shared `filterAndPage` helper; the action ID, transaction hash, and type matcher is unchanged.
- Follow-ups are tracked in #723, #724, and #725. Cardanoscan’s camel-case `govAction` detail route was verified on mainnet, preprod, and preview.
- Focused governance API race tests, the full UI Go suite, `go vet`, the wallet build, and golangci-lint pass.

chrisguiney (approved):

PR #695, opened 2026-08-17
fix(ui): install a Buffer polyfill so the wallet renders

wolf31o2:
## The problem

The wallet renders a **blank page**. `#root` has zero children — React never mounts.

`@ledgerhq/hw-transport-webhid` is imported at module scope from `src/hw/index.ts:18` and reaches for the Node `Buffer` global while it is being evaluated. Browsers have no such global, so the module throws:

```
ReferenceError: Buffer is not defined
```

Because the import is static and reachable from the initial bundle, that abort takes down startup entirely rather than just the Ledger flow.

This reproduces against a **production build**, not only in dev.

## The fix

**1. Install `Buffer` from a dedicated polyfills module, imported first in `main.tsx`.**

It has to be its own module rather than an assignment in `main.tsx`: ES module evaluation runs every import of a module before that module's own body, so `globalThis.Buffer = …` written directly in `main.tsx` would land long after `./app` and the hardware modules beneath it had already been evaluated and thrown.

**2. Load the WebHID transport with a dynamic import.**

`trezor.ts:143` and `keystone.ts:425` already load their transports this way; Ledger was the odd one out. Both of its runtime uses are inside the already-`async` `connectLedger()`, so this is a contained change. It keeps the transport out of the initial bundle — a user who never connects a Ledger never pays for it, and a fault in it can no longer take down startup.

## Why the existing tests missed it

vitest runs on Node, where `Buffer` is a real global. Every runtime assertion passes in the test environment and the failure appears only in a browser — which is why 664 green tests sat alongside a wallet that would not open.

The added tests therefore assert the *structural* properties that keep the browser working: that `./polyfills` is the first import in `main.tsx`, and that the transport has no module-scope import.

## Verification

Built and served the production bundle, then loaded it:

| | `#root` children | `.app` present |
|---|---|---|
| before | `0` | no |
| after | `1` | yes |

`npx tsc -b` clean · 664 tests pass · `eslint` clean.

---
## Summary by cubic
Fixes the blank wallet page by installing a browser `Buffer` polyfill and lazy-loading the Ledger WebHID transport. Previously, a static import of `@ledgerhq/hw-transport-webhid` accessed `Buffer` during module evaluation and crashed startup; now the app mounts and Ledger loads only when needed.

**Review notes**
- Adds `buffer` dependency and `src/polyfills.ts`; `main.tsx` imports it first to install `globalThis.Buffer`.
- Loads `@ledgerhq/hw-transport-webhid` with a dynamic import inside `connectLedger()` to keep it out of the initial bundle.
- Adds source-level tests enforcing polyfills-first import order and no module-scope import of the Ledger transport.
- Required: keep `import "./polyfills"` as the first import in `main.tsx` and avoid top-level imports of `@ledgerhq/hw-transport-webhid`.

<sup>Written for commit 2ec788201caef8ef940131905ce215449dd3d04c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/bursa/pull/695?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ui/web/package.json
+++ b/ui/web/package.json
@@ -22,6 +22,7 @@
     "@zxing/browser": "^0.1.5",
     "@zxing/library": "^0.21.0",
     "bech32": "^1.1.4",
+    "buffer": "^6.0.3",
     "qrcode.react": "^4.2.0",
     "react": "^19.0.0",
     "react-dom": "^19.0.0"
--- a/ui/web/src/hw/ledger.ts
+++ b/ui/web/src/hw/ledger.ts
@@ -10,7 +10,6 @@
  * implementation; the parity vector is exercised in ledger.test.ts.
  */
 
-import TransportWebHID from "@ledgerhq/hw-transport-webhid";
 import Ada from "@cardano-foundation/ledgerjs-hw-app-cardano";
 import type {
   AssetGroup,
@@ -153,6 +152,11 @@ export async function connectLedger(): Promise<HardwareSigner> {
     throw new Error("WebHID not available — open this in a Chromium browser");
   }
 
+  // Loaded on demand, as trezor.ts and keystone.ts load their transports: this
+  // keeps the WebHID transport (and the Node-shaped globals it reaches for) out
+  // of the initial bundle, so a user who never touches a Ledger never pays for
+  // it and a fault in it cannot take down app startup.
+  const { default: TransportWebHID } = await import("@ledgerhq/hw-transport-webhid");
   const transport = await TransportWebHID.create();
   const cardano = new Ada(transport);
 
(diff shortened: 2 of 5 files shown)


wolf31o2:
Superseded by #697, which consolidates this with the other wallet UI work into a single review (23 files, under the cubic limit).

PR #696, opened 2026-08-17
fix(ui): icon-only copy control, readable digits, a table that fits

wolf31o2:
Interface cleanup found by actually running the wallet and looking at it.

> Depends on #695 to render at all — the wallet is currently a blank page.

## Copy is an icon now

A copy control sits beside almost every address, hash and CBOR blob in the wallet — **39 of them**. Every one read `Copy`, which turned dense readouts into a wall of repeated words.

In the Activity table the text did not even fit its column: the button wrapped character-by-character into a vertical `C o p y` and tripled every row's height, at an ordinary 1440px desktop width.

It is now a glyph that swaps to a check on success, square by construction so it cannot wrap. All 39 call sites change through the one component.

Two details worth flagging for review:

- **Accessible name.** Most call sites pass no label. For an icon-only control that would mean *no accessible name at all*, so it falls back to `Copy to clipboard`.
- **State.** The copied state rides on the button's own accessible name, re-announced on the focused control after the click. A live region per control would have put 39 of them in the document competing to be "the" status — which broke four unrelated `findByRole("status")` queries when tried.

## The transaction table fits now

Cells no longer wrap: hashes, addresses, amounts and timestamps are atomic tokens that mean nothing broken mid-way, and `.table-wrap` already scrolls horizontally for a table that outgrows its column.

That alone made rows single-line but pushed the table past its container, hiding the **Details** action off-screen. Fee and block height are drill-down facts, not scanning facts, and the detail drawer already shows both in full — so they are out of the list.

| | before | after |
|---|---|---|
| row height @1440px | 185px | 56px |
| horizontal overflow | — | none |

## Digits are grouped

Live stake read `63120000000000`. Digits were already tabular and monospaced; grouping is the remaining half of making them readable. Zero-decimal tokens (SNEK, HOSKY) group too — they have no fractional part to scale, but their whole-unit counts are the largest numbers on screen.

## One unit spelling

`Staking.tsx` was the only file using `₳` (U+20B3); every other screen writes `ADA`. It renders as a bare `A` wherever the font stack has no glyph for it — which is what it did on the machine these screenshots came from.

## Verification

`npx tsc -b` clean · 668 tests pass · `eslint` clean (one pre-existing warning in `connectorNotify.ts`, untouched).

Four tests that asserted the old contract were retargeted rather than deleted: block height and fee are now asserted *absent* from the list, and the ordering test identifies rows by direction instead of by the removed block column.

---
## Summary by cubic
Makes the wallet UI readable and compact by replacing text copy buttons with an icon-only control, grouping digits in amounts, and trimming the Activity table so rows fit on one line. The list no longer shows Fee or Block, copy state is announced via the button’s accessible name, and the unit spells “ADA” instead of `₳`.

- Copy control: `CopyButton` is now icon-only, swaps to a check on success, and defaults its accessible name to “Copy to clipboard” when no label is provided. State is exposed via `aria-label`/`title`, not visible text, and a new `.btn-icon` style keeps the control square and non-wrapping across 39 call sites.
- Activity table: cells are no-wrap and middle-aligned; Fee and Block are removed from the list and remain in the detail drawer and CSV. This prevents horizontal overflow and keeps the Details action in view.
- Formatting: `formatAda` and `formatTokenQuantity` insert thousands separators in integer parts. Zero- or negative-decimal tokens now show grouped whole-unit counts.
- Unit spelling: Staking switches from `₳` to “ADA” to match other screens and avoid missing glyphs.

**Migration**
- If you relied on Fee or Block in the list, use the drawer or CSV export.
- Provide an `aria-label` to `CopyButton` when the copied value needs a specific accessible name; otherwise it falls back to “Copy to clipboard.”
- Expect comma-grouped amounts in UI and tests; update assertions that matched ungrouped numbers.
- Depends on #695 for the wallet to render.

<sup>Written for commit 0b760927908f6f1b8390acee620e2f4e7f39356a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/bursa/pull/696?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ui/web/src/components/CopyButton.tsx
+++ b/ui/web/src/components/CopyButton.tsx
@@ -1,9 +1,30 @@
-import { useState } from "react";
+import { useEffect, useRef, useState } from "react";
 
 interface CopyButtonProps {
   value: string;
   ariaLabel?: string;
   "aria-label"?: string;
+}
+
+// Icon-only: a copy control sits beside almost every address, hash and CBOR
+// blob in the wallet, and 39 buttons reading "Copy" turned dense readouts into
+// a wall of repeated words. The glyph carries the meaning; the label carries
+// the accessible name.
+function CopyGlyph() {
+  return (
+    <svg viewBox="0 0 16 16" width="14" height="14" aria-hidden="true" focusable="false">
+      <rect x="5.5" y="5.5" width="8" height="8" rx="1.5" fill="none" stroke="currentColor" strokeWidth="1.4" />
+      <path d="M10.5 3.5A1.5 1.5 0 0 0 9 2H3.5A1.5 1.5 0 0 0 2 3.5V9a1.5 1.5 0 0 0 1.5 1.5" fill="none" stroke="currentColor" strokeWidth="1.4" strokeLinecap="round" />
+    </svg>
+  );
+}
+
+function CheckGlyph() {
+  return (
+    <svg viewBox="0 0 16 16" width="14" height="14" aria-hidden="true" focusable="false">
+      <path d="M3 8.5l3.2 3.2L13 5" fill="none" stroke="currentColor" strokeWidth="1.8" strokeLinecap="round" strokeLinejoin="round" />
+    </svg>
+  );
 }
 
 export function CopyButton({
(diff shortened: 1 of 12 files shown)


wolf31o2:
Superseded by #697, which consolidates this with the other wallet UI work into a single review (23 files, under the cubic limit).

PR #697, opened 2026-08-17
fix(ui): make the wallet render, and clean up what it renders

wolf31o2:
Consolidates #695 and #696 and adds the first step of the navigation rework, so this is one review instead of three. 23 files.

Found by building the wallet, running it against a node, and looking at every screen.

---

## 1. The wallet did not render at all

`#root` had zero children. React never mounted.

`@ledgerhq/hw-transport-webhid` is imported at module scope from `src/hw/index.ts:18` and reaches for the Node `Buffer` global while being evaluated. Browsers have no such global, so it threw `ReferenceError: Buffer is not defined` — and because the import is static and reachable from the initial bundle, that took down startup rather than just the Ledger flow. Reproduced against a **production build**, not only dev.

Fixed by installing `Buffer` from a dedicated polyfills module imported first in `main.tsx`. It must be its own module: every import of a module is evaluated before that module's own body, so `globalThis.Buffer = …` written directly in `main.tsx` would land long after the hardware modules had already thrown.

The transport is also now a dynamic import, as `trezor.ts` and `keystone.ts` already do — Ledger was the odd one out. Both its runtime uses are inside the already-`async` `connectLedger()`.

**Why 664 green tests missed it:** vitest runs on Node, where `Buffer` is a real global, so the failure only ever appears in a browser. The added tests assert the structural properties instead — polyfills imported first, no module-scope import of the transport.

| | `#root` children | `.app` present |
|---|---|---|
| before | `0` | no |
| after | `1` | yes |

## 2. Copy is an icon

A copy control sits beside almost every address, hash and CBOR blob — **39 of them**, each reading `Copy`. In the Activity table the text did not even fit its column: the button wrapped character-by-character into a vertical `C o p y` and tripled every row's height at an ordinary 1440px width.

Now a glyph that swaps to a check on success, square by construction so it cannot wrap. Two review-worthy details:

- Most call sites pass no label. For an icon-only control that means *no accessible name at all*, so it falls back to `Copy to clipboard`.
- The copied state rides on the button's own accessible name, re-announced on the focused control. A live region per control would put 39 of them competing to be "the" status — which broke four unrelated `findByRole("status")` queries when tried.

## 3. The transaction table fits

Cells no longer wrap — hashes, addresses, amounts and timestamps are atomic tokens that mean nothing broken mid-way, and `.table-wrap` already scrolls horizontally.

That alone pushed the table past its container and hid the **Details** action off-screen, so fee and block height are out of the list: they are drill-down facts, and the detail drawer already shows both in full.

| | before | after |
|---|---|---|
| row height @1440px | 185px | 56px |
| horizontal overflow | — | none |

<table>
<tr><th width="50%">before</th><th width="50%">after</th></tr>
<tr>
<td>Activity before: Copy buttons wrapped one letter per line, rows three lines tall</td>
<td>Activity after: single-line rows, icon copy control, Details visible</td>
</tr>
</table>

## 4. Digits are grouped, and ADA is spelled one way

Live stake read `63120000000000`. Digits were already tabular and monospaced; grouping is the remaining half. Zero-decimal tokens group too — no fractional part to scale, but the largest numbers on screen.

`Staking.tsx` was the only file using `₳` (U+20B3); every other screen writes `ADA`. It renders as a bare `A` wherever the font stack has no glyph for it.

## 5. Staking, Rewards and Pools become one Stake screen

Three top-level destinations answering one question — what is my ADA doing and where should it go — that did not connect. The directory's own help text told you to *copy a pool ID and paste it into the delegation screen*, with the clipboard as the integration layer between two halves of one task.

They are now tabs of one destination, and **each pool row has a Delegate action** that hands the pool to the delegation form and switches to it. The pool-ID draft is owned by Stake, so browsing the directory mid-compose and coming back does not lose what you typed.

Gating moved with it: the screen needs only a queryable node, so a read-only wallet still gets rewards and the directory, and the Delegation tab explains itself rather than the whole destination vanishing.

`#/staking`, `#/rewards` and `#/pools` still resolve, each opening its matching tab, so existing links keep working.

`Tabs` is extracted from Operate — the only correct ARIA tablist in the app — into a shared component, now with Home/End as well as arrow keys. Operate uses it too, so the wiring exists once instead of twice.

<table>
<tr><th width="50%">before</th><th width="50%">after</th></tr>
<tr>
<td>Stake Pools before: a separate screen, ungrouped digits, Copy only</td>
<td>Pools as a tab of Stake, grouped digits, a Delegate action per row</td>
</tr>
</table>

Choosing a pool lands on the delegation form with it filled in — for a wallet that already delegates too, which is where the first attempt got this wrong:

Delegation tab: current status above, set-up form below with the chosen pool prefilled

**Nav: 17 entries → 15.** More consolidation follows in the navigation rework; this is the first step.

---

## 6. Found by clicking through every screen

I walked all fifteen screens and pressed every control against a mock node. Most of it behaves: send previews and submits, staking withdraw and delegate preview correctly, sign/verify round-trip, contacts, import-transaction decode, the five Operate tabs, settings toggles persist, wallet switching gates exactly the right screens for a hardware wallet, lock returns to the unlock screen, and the mobile drawer navigates and closes. These came out of it.

**A screen that throws no longer takes the wallet with it.** Twice during the pass a single unexpected field in one API response unmounted the entire React tree — blank white page, vault still unlocked, no indication of what happened, no way back but a reload. The node is the source of truth and its responses are not part of this codebase, so a field arriving absent or misshapen is a case to survive, not to assume away.

The boundary wraps the screen area rather than the shell, so the nav survives and you can navigate out; it clears on navigation. It says funds are unaffected, because on a wallet that is the first question a crash raises.

Recoverable error panel with the nav still present

**`/diagnostics` was missing from the dev-server proxy,** so that screen was broken under `npm run dev` — the request fell through to the SPA fallback and the screen reported a JSON parse error at the user.

**Long values overflowed their card on narrow viewports.** A pool ID in a `dl-row` was clipped at the card edge with no ellipsis and no way to read the rest. They wrap now instead of truncating, since those rows carry no copy control and truncating would lose characters outright.

Delegated pool ID wrapping inside its card on a 390px viewport

**"Spending password" was rendered up to three times on one screen** — card title, field label, and an identical placeholder. The duplicate placeholders are gone and the staking card is retitled; the placeholder stays in the two multi-sig fields where it is the only visible affordance or says something the label does not.

One thing deliberately left alone: the Unlock button submits with an empty password rather than being disabled. The backend rejects it, so this is a nicety rather than a hole, and it is not this PR's subject.

## Verification

`npx tsc -b` clean · **678 tests pass** · `eslint` clean (one pre-existing warning in `connectorNotify.ts`, untouched).

Screens were driven against a mock node to check them visually; the pool directory, activity table, portfolio, receive and staking screens were all re-checked after each change. That is how the delegate-handoff bug in the last commit was found — it passed its unit test and still did the wrong thing on screen.

Screenshots are hosted on the `assets/screenshots` branch so the binaries stay off `main`.

## Summary by CodeRabbit

* **New Features**
  * Added unified Staking navigation with Delegation, Rewards, and Pools tabs.
  * Added direct pool delegation actions and preserved in-progress delegation drafts.
  * Added recoverable error messages with retry and reload options.
  * Improved keyboard navigation and accessibility across tabs and copy controls.
* **Improvements**
  * Added thousands separators to ADA and token amounts.
  * Simplified transaction activity displays and improved CSV formatting.
  * Standardized staking currency labels to “ADA.”
* **Bug Fixes**
  * Improved hardware-wallet startup reliability in supported browsers.

### DIFF
--- a/ui/web/package.json
+++ b/ui/web/package.json
@@ -22,6 +22,7 @@
     "@zxing/browser": "^0.1.5",
     "@zxing/library": "^0.21.0",
     "bech32": "^1.1.4",
+    "buffer": "^6.0.3",
     "qrcode.react": "^4.2.0",
     "react": "^19.0.0",
     "react-dom": "^19.0.0"
--- a/ui/web/src/app.tsx
+++ b/ui/web/src/app.tsx
@@ -21,9 +21,7 @@ import { Activity } from "./screens/Activity";
 import { Send } from "./screens/Send";
 import { Swap } from "./screens/Swap";
 import { Contacts } from "./screens/Contacts";
-import { Staking } from "./screens/Staking";
-import { RewardHistory } from "./screens/RewardHistory";
-import { PoolDirectory } from "./screens/PoolDirectory";
+import { Stake } from "./screens/Stake";
 import { SignMessage } from "./screens/SignMessage";
 import { VerifyMessage } from "./screens/VerifyMessage";
 import { Offline } from "./screens/Offline";
@@ -61,6 +59,15 @@ const ROUTES = new Map<string, () => ReactElement>([
   ["diagnostics", Diagnostics],
 ]);
 
+// Stake merged three screens; the old routes stay valid and open the matching
+// tab so existing bookmarks and links keep working.
+const STAKE_ROUTES = new Map<string, "delegation" | "rewards" | "pools">([
+  ["stake", "delegation"],
+  ["staking", "delegation"],
+  ["rewards", "rewards"],
+  ["pools", "pools"],
+]);
+
 const NAV: { key: string; label: string }[] = [
   { key: "portfolio", label: "Portfolio" },
   { key: "receive", label: "Receive" },
@@ -68,9 +75,7 @@ const NAV: { key: string; label: string }[] = [
   { key: "send", label: "Send" },
   { key: "swap", label: "Swap" },
   { key: "contacts", label: "Contacts" },
-  { key: "staking", label: "Staking" },
-  { key: "rewards", label: "Rewards" },
-  { key: "pools", label: "Stake Pools" },
+  { key: "stake", label: "Stake" },
   { key: "sign", label: "Sign" },
   { key: "verify", label: "Verify" },
   { key: "offline", label: "Offline" },
@@ -283,9 +288,9 @@ export function App() {
     if (route === "settings") activeRoute = "settings";
     else if (route === "send" && canSend) activeRoute = "send";
     else if (route === "swap" && canSwap) activeRoute = "swap";
-    else if (route === "staking" && canStake) activeRoute = "staking";
-    else if (route === "rewards") activeRoute = "rewards";
-    else if (route === "pools" && canQueryNode) activeRoute = "pools";
+    // The legacy per-screen routes now resolve to tabs of the merged screen,
+    // so an old bookmark or a link in the wild still highlights Stake.
+    else if (STAKE_ROUTES.has(route) && canQueryNode) activeRoute = "stake";
     else if (route === "sign" && canSign) activeRoute = "sign";
     else if (route === "verify") activeRoute = "verify";
     else if (route === "offline" && canSign) activeRoute = "offline";
@@ -334,30 +339,22 @@ export function App() {
     // Guard deep-links (#/swap): DEX quotes need a queryable mainnet node, so
     // fall back to Portfolio while the node or active wallet cannot support it.
     content = <Portfolio />;
-  } else if (route === "staking") {
-    // Staking/governance is gated like send: a synced node AND a
-    // spending-enabled wallet. A read-only or unsynced wallet falls back to
-    // Portfolio.
-    content = canStake ? (
-      <Staking
+  } else if (STAKE_ROUTES.has(route)) {
+    // Delegation, rewards and the pool directory are one screen. It needs only
+    // a queryable node, so read-only wallets still get rewards and pools; the
+    // Delegation tab explains itself when this wallet cannot sign a
+    // certificate, rather than the whole screen vanishing.
+    content = canQueryNode ? (
+      <Stake
         network={activeWallet.network}
         isHardware={activeWallet.type === "hardware"}
         walletId={activeWallet.id}
+        canDelegate={canStake}
+        initialTab={STAKE_ROUTES.get(route)}
       />
     ) : (
       <Portfolio />
     );
-  } else if (route === "rewards") {
-    // Read-only per-epoch reward history for the active wallet's stake
-    // address. Node-local read (no signing, no spend), so it is available to
-    // any active wallet; the pool explorer links need the wallet's network.
-    content = <RewardHistory network={activeWallet.network} />;
-  } else if (route === "pools") {
-    // Read-only stake-pool directory: browse/search pools the node has indexed.
-    // Needs only a queryable node (not a full sync, no spending), so it works
-    // for any active wallet — including read-only ones. Falls back to Portfolio
-    // while the node cannot serve queries.
-    content = canQueryNode ? <PoolDirectory network={activeWallet.network} /> : <Portfolio />;
   } else if (route === "sign") {
     content = canSign ? <SignMessage account={toAccount(activeWallet)} /> : <Portfolio />;
   } else if (route === "verify") {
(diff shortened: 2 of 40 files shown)


wolf31o2:
Thanks — this was a good review. All eleven addressed in `da69d1b`, and two of them were worse than reported.

**The CSS rename was wrong twice over (P2 + P3).** Correct that `.operate-tab` still has three consumers — Registration, Retirement and OperationalCert use it for their seed/air-gap mode toggles, and deleting it left them as unstyled browser buttons. The part not mentioned: **`.tab` was already defined at `global.css:1184`** for Offline's underline-style `.tab-bar`, so my block created a *second* `.tab` ruleset that silently restyled that screen. The shared component now uses `.tabset` / `.tabset-tab`, colliding with neither, and `.operate-tab` is restored alongside it. Verified by computed style: Offline transparent/0px-radius underline, Operate gold pill, tabset its own.

**P1 error-panel reassurance — agreed and important.** "Nothing was submitted" is not knowable; a render fault can land after a submission and a false all-clear invites a duplicate send. It now says the keys are safe (true, and the first question a crash raises on a wallet) and tells the user to check Activity before retrying.

**P2 non-Error throws — valid.** A child can throw a string, `null`, or `undefined`, and reading `.message` off that re-threw *inside the fallback*, restoring the blank page the boundary exists to prevent. The caught value is normalised now, with tests for `throw "boom"` and `throw null`.

**P2 `resetKey`** — correct, `activeRoute` collapses stake/staking/rewards/pools to one key. Keys on the raw route now.

**P2 Stake `initialTab`** and **P2 unconditional `onDelegate`** — both valid, both fixed, both now have tests.

**P2 CopyButton label** — valid, and the better call. With a copy control beside every address and hash, a bare "Copied" leaves a screen-reader user unable to tell which one. Now `${label} (copied)`.

**P2/P3 `formatTokenQuantity`** — agreed. Out-of-range decimals return unchanged; grouping applies only at `decimals === 0`. Docstring updated to match.

**P3 `.sr-only` dead** and **P3 copied-hover specificity** — both correct, both fixed.

**P1 `composeSignal`** — already fixed in `8222a96`, before this review landed; you were reviewing the commit before it. Worth noting the root cause you identified was exactly right (the tab unmounts `Staking`, so the remount initialises `seenSignal` to the already-incremented nonce). I took the other branch of your suggestion — an explicit `startInCompose` prop cleared once consumed, rather than `composeSignal > 0` lazy state — because the nonce stays above zero after the flow completes, so lazy state would re-enter compose on every later visit to the tab.

682 tests pass, `tsc` and `eslint` clean.

wolf31o2:
Second round addressed in `943648e`. Two of these were more than nits.

**CSV export (P2) — the serious one, and correct.** `transactionsToCsv` formats `net_ada`/`fee_ada` with `formatAda`, so grouping turned a 1,500 ADA transaction into `1,500`, which a consumer reads as quoted text or splits across columns. That is data corruption in an export, introduced by a display change — exactly the kind of thing that is invisible until someone opens the file in a spreadsheet. Display and machine output are now separate: `formatAda` groups, `formatAdaPlain` does not, the exporter uses the latter. `transactionsToCsv` is exported so the column contract is asserted directly rather than through a `Blob` jsdom cannot read.

**Rewards gating (P2) — you were right that it contradicted the stated intent, and the intent was the correct one.** Not intentional narrowing: the old `#/rewards` had no node gate, so merging the screens quietly took reward history away from a read-only wallet with a stopped node. The screen is no longer gated as a whole; each tab states its own requirement (Delegation on signing capability, Pools on node queryability), which is what the comment claimed all along.

**Compose state (P2) — valid, and lifting only `poolId` was half a fix.** The whole draft is lifted now, *and* whether the form is open: an already-delegating wallet returning from the directory landed on its status panel with the half-filled form hidden behind "Change delegation", which loses the user's place just as surely as losing the fields. Verified in a browser as well as in tests — pool, vote target and the open form all survive the round trip.

**`addingWallet` in resetKey (P2)** — correct, it swaps content without changing the route. Fixed.

**Boundary transient clear (P3)** — correct and subtle. The boundary now records which key an error was captured under and clears only errors predating the change, so a destination that throws in the same commit is no longer discarded, remounted, thrown again and logged twice.

**Copy control labels (P2)** — fair, and it was pre-existing rather than introduced, but the icon change is a good moment to fix it: 27 call sites passed no label. All 27 now carry a specific one.

**Live region (P2)** — agreed that a name change on an already-focused control is not reliably announced. Reconciling this with your earlier point about competing status nodes: the region renders *only while copied*, so at most one exists at a time and the document never holds 39 empty ones.

**`groupDigits` splitting the JSDoc (P3)** — correct, moved.

686 tests pass, `tsc` and `eslint` clean. 40 files, still inside the cubic limit.

arepala-uml (approved):

PR #666, opened 2026-08-04
feat(ui): multi-account support (BIP44 account switching)

wolf31o2:
One wallet (seed) can now expose multiple BIP44 accounts (account index 0,1,2,…), each with its own addresses/balance, with view + switch + add. Single-account wallets are unchanged (default account index 0).

## Derivation
- CIP-1852 account index is parameterized: `wallet.DeriveAccount`, `wallet.DeriveAccountFromMnemonicBytes`, `wallet.AccountXpubForIndexFromMnemonicBytes` (existing `Derive`/`DeriveFromMnemonicBytes`/`AccountXpub` delegate to account 0). `Account.AccountIndex` is now set for mnemonic-derived accounts.
- Signing derives at the bound account's index: `spend` and `poolops` use `GetAccountKey(root, acct.AccountIndex)` instead of a hardcoded 0.

## Vault (backward compatible)
- `WalletMeta.Accounts []*wallet.Account` (additive, `omitempty`) holds every derived account; legacy records (nil) synthesize a single-element list from the canonical index-0 `Account`.
- Selected account index per wallet persists in a new cleartext, non-secret `envelope.active_accounts` map (default 0), kept out of the encrypted index so a switch needs no vault password / no re-seal.
- `AddAccount(id, vaultPassword, spendPassword, accountIndex, windowN)` derives + stores a new account (seed to derive the hardened account, vault password to re-seal the index). `SelectAccount(id, accountIndex)` sets the active index (no password). `ActiveAccountIndexFor(id)`.
- On-disk format version unchanged; existing vaults load and default to account 0. Hardware/read-only wallets reject `AddAccount` (`ErrNoSeed`).

## Endpoints
- `GET /wallet/accounts` — active wallet's accounts: index, label, stake address, first receive address, active flag, best-effort node-local balance summary.
- `POST /wallet/account/select` `{account_index}` — set the active account index and rebind read/spend/pool services.
- `POST /wallet/accounts` `{account_index, vault_password, spend_password}` — derive + add a new account.
- `bindActive` binds the wallet's active account; `walletView` carries `accounts` + `active_account_index`, and `stake_address`/`addresses` reflect the active account.

## Frontend
- `AccountSwitcher` (list, switch, add) nested under the active wallet in `WalletSwitcher` (desktop sidebar + mobile drawer). Add-account is offered only for seed-backed wallets.
[...]

PR #704, opened 2026-08-17
feat(ui): fold actions and housekeeping out of the nav (15 → 9)

wolf31o2:
Second step of the navigation rework, after #697. **Nav goes from 15 entries to 9, and fits a laptop viewport without scrolling for the first time.**

Nothing is removed — things move to where they belong.

<table>
<tr><th width="50%">before — 15 entries</th><th width="50%">after — 9, with Send/Receive on the balance</th></tr>
<tr>
<td>Sidebar with fifteen flat entries</td>
<td>Nine-entry sidebar, Send and Receive as buttons under the balance</td>
</tr>
</table>

## Send and Receive are actions, not places

They now sit on the balance card they act on — where you already are when you decide to move funds — rather than costing two entries in a list you have to scan. A primary button on the screen you land on is *more* prominent than a sidebar row, not less; it is also how Lace, Yoroi and Eternl handle it.

They keep their routes. Deep links, the connector, and anything else pointing at `#/send` still work; only the way in moved.

## The address book, message tools and diagnostics are Settings panels

Four more top-level entries, each rarely used, none of them somewhere you go to *do wallet work*. Settings gains the shared `Tabs` component from #697 rather than growing into a dozen stacked cards.

Settings with General, Address book, Tools and Diagnostics tabs; Tools showing Sign and Verify

Signing needs the wallet's seed, so the Tools panel explains itself for a hardware or watch-only wallet instead of presenting a form that cannot work. Verification is pure crypto and stays available to any wallet.

## Old links keep working

Every route these screens used to own still resolves and opens the matching panel, and the nav highlight follows:

| route | opens | highlights |
|---|---|---|
| `#/contacts` | Settings → Address book | Settings |
| `#/sign`, `#/verify` | Settings → Tools | Settings |
| `#/diagnostics` | Settings → Diagnostics | Settings |
| `#/send`, `#/receive` | the screen itself | Portfolio |

## Rebased onto multi-account (#666)

That PR's account switcher sits inside the wallet list, and the shorter nav is what gives it room: even with two wallets and two accounts the sidebar no longer scrolls.

Two small fixes to it are folded in here, since this PR is already reshaping that sidebar:

- Its balances used `₳` (U+20B3), which renders as a bare **A** wherever the font stack has no glyph for it — the same defect #697 removed from `Staking`. Spelled `ADA`, as everywhere else in the wallet.
- The account label now truncates instead of wrapping "Account 0" onto two lines in a 232px sidebar.

## Verification

`tsc` clean · **700 tests pass** · `eslint` clean (one pre-existing warning in `connectorNotify.ts`, untouched). 7 files.

Driven in a browser against a mock node: all nine nav entries, all four Settings panels, every legacy route above, and mobile at 390px (drawer lists 9, tab strip fits, no horizontal overflow). After the rebase, also that switching account actually switches it, not merely that the switcher renders.

Cubic caught a P0 on the first push — Portfolio resolved through the props-less `ROUTES` map, so `canSend` defaulted false and the Send button was permanently disabled, which with Send out of the nav made sending unreachable. Fixed, with a test I confirmed fails against the broken commit. Worth naming the gap: every check I had run was structural (does it render, does the route resolve, does the nav highlight) and none of them pressed the button.

## What follows

The last step of the agreed rework is structural rather than cosmetic: multi-sig accounts move into the wallet switcher (`WalletType` already has `multi_signature`), Import Tx and Offline become modes of the send/sign flow, Pool Ops appears only when enabled, and a ⌘K palette with a visible trigger covers everything demoted. That takes the nav to 5.

## Summary by CodeRabbit

* **New Features**
  * Added Send and Receive actions to the Portfolio screen.
  * Disabled sending for wallets that cannot spend.
  * Consolidated Contacts, signing, verification, and Diagnostics into Settings tabs.
  * Preserved access to legacy routes and Settings panel mappings.
  * Improved sidebar navigation to show destination screens more clearly.

* **UI Improvements**
  * Balance displays now use the `ADA` label.
  * Long account names are truncated neatly.
  * Portfolio actions have improved spacing and layout.

### DIFF
--- a/ui/web/src/app.test.tsx
+++ b/ui/web/src/app.test.tsx
@@ -749,19 +749,50 @@ test("a wallet that cannot spend gets a disabled Send, not a dead end", async ()
 
 test("the routes the absorbed screens used to own still resolve", async () => {
   // Old links and bookmarks must not break just because the screen moved.
-  for (const [hash, expected] of [
-    ["#/contacts", "Settings"],
-    ["#/sign", "Settings"],
-    ["#/verify", "Settings"],
-    ["#/diagnostics", "Settings"],
-    ["#/receive", "Portfolio"],
+  // Asserting the nav highlight alone is not enough: every one of these
+  // highlights Settings, so a mis-mapped SETTINGS_ROUTES entry would sail
+  // through while opening the wrong panel. Check the panel that opened.
+  for (const [hash, navEntry, panel] of [
+    ["#/contacts", "Settings", "Address book"],
+    ["#/sign", "Settings", "Tools"],
+    ["#/verify", "Settings", "Tools"],
+    ["#/diagnostics", "Settings", "Diagnostics"],
+    ["#/settings", "Settings", "General"],
   ] as const) {
     window.location.hash = hash;
     const sidebar = await unlockAndGetSidebar(walletA);
-    const active = within(sidebar).getByRole("button", { name: expected });
-    expect(active).toHaveAttribute("aria-current", "page");
+
+    expect(within(sidebar).getByRole("button", { name: navEntry })).toHaveAttribute(
+      "aria-current",
+      "page",
+    );
+    expect(await screen.findByRole("tab", { name: panel })).toHaveAttribute(
+      "aria-selected",
+      "true",
+    );
+
     cleanup();
     vi.restoreAllMocks();
     stubAutoLock(0);
   }
+});
+
+test("#/receive opens Receive while the nav still points at Portfolio", async () => {
+  // Send and Receive are Portfolio actions, so the highlight is deliberate —
+  // but the screen on display is Receive, and the error boundary must say so
+  // rather than blaming Portfolio for a fault in a different screen.
+  vi.spyOn(hooks, "useAddresses").mockReturnValue({
+    data: { receive: ["addr_test1abc"], used: [], next_unused: "addr_test1abc" },
+    error: null,
+    loading: false,
+    refresh: vi.fn(),
+  } as never);
+  window.location.hash = "#/receive";
+  const sidebar = await unlockAndGetSidebar(walletA);
+
+  expect(within(sidebar).getByRole("button", { name: "Portfolio" })).toHaveAttribute(
+    "aria-current",
+    "page",
+  );
+  expect(await screen.findByText(/next unused address/i)).toBeInTheDocument();
 });
--- a/ui/web/src/app.test.tsx
+++ b/ui/web/src/app.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen, waitFor, fireEvent, act, within } from "@testing-library/react";
+import { render, screen, waitFor, fireEvent, act, within, cleanup } from "@testing-library/react";
 import { App } from "./app";
 import * as hooks from "./api/hooks";
 import * as client from "./api/client";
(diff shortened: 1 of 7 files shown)


wolf31o2:
Good catch, and correctly rated P0 — fixed in `f52a126`.

Exactly right on the mechanism: `"portfolio"` is in the `ROUTES` map, the final `else` resolves it as `<Screen />` with no props, and `canSend` defaults to `false`. Since this commit removed Send from the nav, that button is the only way into the send flow, so the net effect was that sending became unreachable. Portfolio is now constructed directly on that path with `canSend` passed.

Worth noting how it got past me: I drove this in a browser and checked the nav count, every legacy route, all four Settings panels and mobile — but I only asserted the Send button *rendered*, never that it was enabled or that clicking it went anywhere. Structure, not behaviour. The tests now assert both, plus that a wallet which cannot spend gets a disabled Send while Receive stays available; I reverted the fix to confirm the new test actually fails against the previous commit rather than passing either way.

692 tests, `tsc` and `eslint` clean.

wolf31o2:
Both addressed in `cf90c1b`.

**Boundary label (`app.tsx:299`)** — right, and it is the same mislabeling the `STAKE_ROUTES` comment claims to avoid. The two things were conflated: `activeRoute` is the *nav highlight*, which deliberately collapses routes onto one entry, and I was reusing it as the *screen name*. Highlighting Portfolio while on Send is intended — Send being a Portfolio action is the whole point of the change — but a fallback reading "the portfolio screen could not be displayed" when Send is what broke points the reader at the wrong screen. The boundary names the route on display now; the highlight is untouched.

**Legacy-route test (`app.test.tsx:762`)** — correct, and the more useful catch. Every one of those routes highlights Settings, so the assertion could not distinguish a correct mapping from a mis-mapped one; the test would have passed with `#/diagnostics` opening General. It asserts the panel that actually opened, which is the claim the PR is making. Added a case for `#/receive` too: the nav points at Portfolio while Receive is what renders, and that pair is now pinned.

Worth naming the pattern, since this is the second time in this branch: the P0 you found earlier was the same shape — I asserted that the Send button *rendered* rather than that it worked, and here I asserted a nav highlight rather than what the user ends up looking at. Structure over outcome, both times.

701 tests, `tsc` and `eslint` clean.

wolf31o2:
Both correct — and the first is a fault in my own fix, not the original code.

**Fallback labels (`app.tsx:320`)** — naming the raw route fixed the collapsed-highlight case and broke the fallback ones. Several branches decline the route they were asked for and render Portfolio instead: `#/swap` on a testnet wallet, the seed-gated screens (`send`, `offline`, `operate`, `import`) for a hardware or watch-only wallet, an unrecognised hash, and the wallet-select card when no wallet is active. In all of those the route names a screen that never appeared. Each branch now sets the label to what it rendered, so the raw route survives only where the route really is what is on display.

**The over-claiming test (`app.test.tsx:780`)** — fair, and the sharper point. The comment asserted something about the error boundary that the test never exercised; it only confirmed Receive rendered. It injects a fault into Receive now and asserts the boundary says "receive", and a second test covers the fallback case by faulting a `#/swap` that renders Portfolio. I re-introduced the bug to confirm both fail against the previous commit rather than passing either way.

That is three findings in this branch with the same root: asserting the thing near the behaviour instead of the behaviour. Worth me being blunt about it — the pattern is mine, not a series of unrelated slips.

703 tests, `tsc` and `eslint` clean.

chrisguiney (approved):

PR #765, opened 2026-08-24
fix(api): protect legacy wallet endpoints

### LINKED ISSUE #742: api: protect legacy server exposure and secret-returning endpoints
The Bursa legacy API must preserve the deployment contract while protecting secret material. The API must retain the wildcard bind required by the shipped Helm chart only for an explicitly configured service, refuse unauthenticated non-loopback exposure, and keep loopback-only operation available; secret-returning wallet creation must not use a cacheable GET or expose mnemonics/private keys without an explicit protected flow; signing, witnessing, data-signing, and address-derivation endpoints must require authentication/authorization; and request decoders must cap bodies.

Acceptance criteria:

- The Helm chart's required wildcard service continues to operate when explicitly configured.
- A non-loopback listener without authentication and the required protection fails startup.
- Secret responses use non-cacheable methods/headers and protected authorization.
- Request body limits and endpoint authorization tests cover all affected handlers and override paths.

wolf31o2:
## Summary

- require TLS plus JWT or JWKS authentication when the legacy API listens beyond loopback
- protect wallet, signing, witness, and address-derivation routes
- use `POST` for wallet creation, prevent caching of secret responses, and cap request bodies
- update Swagger, OpenAPI client documentation, and regression coverage

## Migration

- configure `API_TLS_CERT_FILE` and `API_TLS_KEY_FILE` for non-loopback listeners
- configure exactly one of `API_JWT_SECRET` or `API_JWKS_URL`; issuer and audience remain optional
- call `POST /api/wallet/create` and send `Authorization: Bearer <JWT>` to protected routes

## Validation

- `go test ./...`
- `go test -race ./internal/api ./internal/config`
- `go vet ./...`
- `golangci-lint run ./internal/api/... ./internal/config/...`
- `nilaway ./internal/api/... ./internal/config/...`
- `modernize ./internal/api/... ./internal/config/...`
- `make swagger`
- `(cd openapi && go test ./... && go vet ./...)`
- `git diff --check`

Fixes #742


arepala-uml (approved):
## Review: protect legacy wallet endpoints

**Summary:** Locks down the legacy Bursa API — requires TLS + a single JWT trust source (HS256 secret or JWKS) whenever the listener is non-loopback, wraps wallet/signing/witness/address-derivation routes in bearer-JWT auth, moves `wallet/create` from GET→POST, adds no-store headers on secret responses, and caps request bodies. Docs/OpenAPI client regenerated to match.

**What I checked:**
- `internal/api/api.go` — `buildAPIAuth`, `apiTLSConfigured`, `validateAPIExposure`, `registerAPIHandlers`, and the TLS wiring in `Start`
- `internal/config/config.go` — new `ApiConfig` fields (`TLSCertFile`, `TLSKeyFile`, `JWTSecret`, `JWKSURL`, `JWTIssuer`, `JWTAudience`)
- Reused auth primitives in `internal/signer/api/auth.go` (HS256Validator, JWKSValidator, AuthMiddleware) and `internal/signer.IsLoopbackListenAddress` — pre-existing, already covered by their own tests
- Test additions in `api_test.go` / `config_test.go`, and generated `docs/`, `openapi/` diffs
- `go build ./...` succeeds on the branch

**Findings:**

✅ **Fail-closed exposure check is correct.** Non-loopback + (no auth OR no TLS) refuses to start (`validateAPIExposure`). HS256 secret enforces a 32-byte minimum and is mutually exclusive with JWKS; JWKS enforces HTTPS except on loopback (verified in `auth.go`). Default `ListenAddress` is `127.0.0.1`, so existing loopback deployments are unaffected.

✅ **Route protection matches the stated scope** — wallet CRUD, tx sign/witness, sign/data, address/enumerate are wrapped in `protectAPIHandler`; health/metrics/swagger stay public. Verified via `TestRegisterAPIHandlersProtectsSensitiveRoutes`.

✅ **Test coverage is strong** — auth-builder validation, exposure-matrix (loopback × TLS × auth), an actual TLS-handshake end-to-end test, WWW-Authenticate header checks, and config env/YAML round-trips.

✅ **Generated docs/client are consistent** with the handler changes (POST for wallet/create, `BearerAuth` security, `https` scheme). Some unrelated Conway-governance fields also appear in the regenerated `docs/docs.go` — these already existed in `transaction.go` on `main` and were just missing from previously-stale generated docs; not introduced by this PR.

**Nits (non-blocking):**
- `apiTLSConfigured(cfg)` is computed once inside `validateAPIExposure` and again in `Start` right after — harmless (pure function of `cfg`) but could be returned from the validator to avoid the duplicate call.
- `handleWalletList` doesn't call `setSecretResponseHeaders` like the other wallet handlers — worth confirming that's intentional since it only returns wallet names, not secret material.
- `configureSwaggerScheme` mutates the package-level `docs.SwaggerInfo.Schemes` on every `Start()` call; fine for a single process, but would need care if `Start` is ever called concurrently with different TLS configs in tests.

PR #793, opened 2026-08-27
fix(api): authenticate legacy wallet operations

### LINKED ISSUE #790: security(api): authenticate legacy wallet secret operations
Resolved by [PR #793](https://github.com/blinklabs-io/bursa/pull/793).

The legacy wallet API exposes and deletes GCP-backed wallet secrets without authentication or TLS. An unauthenticated caller can access legitimate clients' wallet key material.\n\nAcceptance criteria:\n- Require authentication and integrity-protected transport for all legacy secret read/delete operations.\n- Preserve an explicit, documented loopback-safe development mode if the existing architecture requires it.\n- Add regression tests proving unauthenticated access is rejected and authorized access remains functional.\n- Update API/security documentation and deployment guidance.\n- Coordinate with the existing Bursa/Helm deployment contract; do not make the Service unreachable as a standalone fix.

wolf31o2:
Fixes #790

## Summary
- document and enforce the wallet administrator 401/403 contract
- avoid counting unchanged wallet updates as writes
- regenerate the nested OpenAPI client and document the Bearer prefix
- add a generated-client wire-header regression test

## Validation
- focused authorization and metric tests
- nested `go test -race -count=1 ./...`
- nested `go vet ./...`

Deployment wiring remains coordinated through helm-charts draft PR #460 / issue #459; this PR does not make the standalone service unreachable.

---
## Summary by cubic
Enforces JWT authentication and wallet-administrator authorization on the persisted wallet list/get/update/delete endpoints when Google Secret Manager persistence and API auth are enabled. These routes were previously unauthenticated; they now return 401 for missing or invalid tokens and 403 when the caller's subject is not in `API_JWT_ADMIN_SUBJECTS`, and an empty admin list fails closed.

**Migration**
- Set `API_JWT_ADMIN_SUBJECTS` (`api.jwt_admin_subjects`) when enabling wallet persistence with API auth; an empty list denies all persisted-wallet operations.

**Additional changes**
- Unchanged wallet updates no longer count as writes.
- Regenerated the OpenAPI client: documents the Bearer auth prefix, adds address and transaction endpoints, regenerates request field tables, guards multipart form encoding, and adds a wire-header regression test.
- Wallet restore requests and `BursaWallet` now include pool cold key fields.

Deployment wiring remains coordinated through the helm-charts draft PR; this PR does not make the standalone service unreachable.

<sup>Written for commit 72c5a317f4c202fc6fdac7a7df30b48701836600. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/bursa/pull/793?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added address building, enumeration, and parsing API operations.
  * Added data signing and verification endpoints.
  * Added transaction assembly, decoding, ID, signing, and witness operations.
  * Added transaction inspection details and wallet pool-cold key support.
  * Added `pool_cold_id` to wallet restoration requests.

* **Security**
  * Wallet storage management now requires authentication and administrator access, returning clear 401 or 403 responses.

* **Documentation**
  * Updated API references and deployment guidance for the new operations, authentication, and configuration requirements.


openapi/client.go
```diff
@@ -323,13 +308,9 @@ func (c *APIClient) prepareRequest(
 	}
 
 	// add form parameters and file if available.
-	if strings.HasPrefix(headerParams["Content-Type"], "multipart/form-data") &&
-		len(formParams) > 0 ||
-		(len(formFiles) > 0) {
+	if strings.HasPrefix(headerParams["Content-Type"], "multipart/form-data") && len(formParams) > 0 || (len(formFiles) > 0) {
```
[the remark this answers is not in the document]
wolf31o2: Fixed in 9a1672b. Multipart encoding now requires a multipart content type for both form fields and files. TestPrepareRequestMultipartRequiresMultipartContentType covers non-multipart file input.

openapi/configuration.go
```diff
@@ -29,6 +29,9 @@ func (c contextKey) String() string {
 }
 
 var (
+	// ContextAPIKeys takes a string apikey as authentication for the request
+	ContextAPIKeys = contextKey("apiKeys")
```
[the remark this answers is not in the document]
wolf31o2: Fixed in 9a1672b. The comment now specifies map[string]APIKey, matching the generated request authentication lookup.

arepala-uml (commented):
Reviewed at head 9a1672b. No merge blockers — the 401/403 authorization contract, the no-op-update counter fix, and the regenerated OpenAPI client are correct and covered by the new tests. Two non-blocking findings below, each reproduced with a standalone test against this head (not committed to the branch).

Also noted but not filed as separate threads (both non-blocking, follow-up material): `legacyWalletStore`/`gcpLegacyWalletStore` (api.go:533) duplicates the existing, currently-unused `internal/storage.Store` abstraction, and the two have already diverged in list-error handling; `registerAPIHandlers`'s variadic `stores ...legacyWalletStore` (api.go:768) only ever reads `stores[0]`, silently dropping anything past the first element.

internal/api/api.go
```diff
@@ -657,6 +747,9 @@ func validateAPIExposure(cfg *config.Config, auth signerapi.Validator) error {
 			cfg.Api.ListenAddress,
 		)
 	}
+	if cfg.Google.Project != "" && cfg.Google.ResourceId != "" && auth != nil && len(cfg.Api.JWTAdminSubjects) == 0 {
```
arepala-uml: Non-blocking: this length check accepts a `JWTAdminSubjects` list made entirely of empty strings. `envconfig` (kelseyhightower/envconfig v1.4.0) parses `API_JWT_ADMIN_SUBJECTS=","` into `["", ""]` — `len(...) == 0` is false, so this guard passes — but `protectWalletStorageHandler` below filters out empty subjects when building its admin set, leaving it empty. The server starts cleanly and then 403s every legitimate admin. Fails closed (no unauthorized access), but it defeats the fail-fast check this line exists to provide.

Reproduced:
```go
cfg.Api.JWTAdminSubjects = []string{"", ""} // what envconfig produces for "API_JWT_ADMIN_SUBJECTS=,"
require.NoError(t, validateAPIExposure(cfg, auth)) // passes
// ...a request carrying a valid, non-empty-subject admin JWT still gets 403
```
Both assertions pass against this head. Suggest validating that at least one non-empty subject is present (e.g. filter/trim during config load, or check after filtering blanks here) rather than just list length.

internal/api/api.go
```diff
@@ -1250,8 +1345,12 @@ func handleWalletUpdate(w http.ResponseWriter, r *http.Request) {
...
-	g := gcp.NewGoogleWallet(req.Name)
-	if err := g.Load(r.Context()); err != nil {
+	updated, err := legacyStore(r).Update(
+		r.Context(),
+		req.Name,
+		req.Description,
+	)
+	if err != nil {
 		logger.Error(
 			"failed to load google wallet",
```
arepala-uml, line 1375: Non-blocking: `gcpLegacyWalletStore.Update` returns both its `Load` error and its `Save` error through the same `error`, so this branch always logs `"failed to load google wallet"` — even when the failure happened in `Save` (e.g. a Secret Manager write-quota or write-permission error). Pre-PR code logged that case distinctly as `"failed to save google wallet"`; that message no longer appears anywhere for this path, which will misdirect debugging toward a read/access problem when it's actually a write problem. The analogous collapse also affects `handleWalletGet`'s `Load`+`PopulateTo` around line 1228 (the old dedicated "failed to convert google wallet" branch is gone).

Reproduced with a log-capturing test: a `legacyWalletStore` double whose `Update` fails with `"secretmanager: write quota exceeded"` produces a log line containing `"failed to load google wallet"` and the write-quota text, and never `"failed to save google wallet"`. All three assertions pass against this head.

Consider having the store return a typed/wrapped error (or split `Update` into distinct load/save-reporting) so the handler can log the correct phase.

internal/api/api.go
```diff
@@ -628,6 +695,29 @@ func protectAPIHandler(auth signerapi.Validator, next http.Handler) http.Handler
 	)
 }
 
+func protectWalletStorageHandler(
+	auth signerapi.Validator,
+	adminSubjects []string,
+	next http.Handler,
+) http.Handler {
+	if auth == nil {
```
arepala-uml, line 708: **Request changes — GCP wallet routes fail open without JWT authentication**

Location:

- `internal/api/api.go`, function `protectWalletStorageHandler`, around line 698
- `internal/api/api.go`, function `validateAPIExposure`, around line 750

`protectWalletStorageHandler` returns the original unprotected handler when `auth == nil`:

```go
if auth == nil {
    return next
}
```

At the same time, `validateAPIExposure` requires `jwt_admin_subjects` only when authentication is already configured:

```go
if cfg.Google.Project != "" &&
    cfg.Google.ResourceId != "" &&
    auth != nil &&
    len(cfg.Api.JWTAdminSubjects) == 0 {
    // ...
}
```

This leaves a fail-open configuration: when GCP wallet storage is enabled on a loopback listener without `jwt_secret` or `jwks_url`, the `/api/wallet/list`, `/api/wallet/get`, `/api/wallet/update`, and `/api/wallet/delete` routes are registered without authentication. In particular, `/api/wallet/get` can return sensitive wallet material, including the mnemonic.

Loopback should not be treated as authorization because other local processes can access the listener. This also contradicts the PR documentation describing these routes as authenticated administrator-only operations.

Please require JWT authentication and at least one non-empty administrator subject whenever GCP wallet storage is enabled. `protectWalletStorageHandler` should also fail closed when called without an authenticator.

A regression test should cover the complete handler behavior:

```go
func TestGCPWalletRoutesFailClosedWithoutAuth(t *testing.T) {
    cfg := &config.Config{
        Google: config.GoogleConfig{
            Project:    "project",
            ResourceId: "kms-resource",
        },
    }

    store := &fakeLegacyWalletStore{
        wallets: map[string]*bursa.Wallet{
            "secret": {
                Mnemonic: "sensitive mnemonic",
            },
        },
    }

    mux := http.NewServeMux()
    registerAPIHandlers(mux, cfg, nil, store)

    req := httptest.NewRequest(
        http.MethodPost,
        "/api/wallet/get",
        strings.NewReader(`{"name":"secret"}`),
    )
    rec := httptest.NewRecorder()
    mux.ServeHTTP(rec, req)

    require.Equal(t, http.StatusUnauthorized, rec.Code)
    require.NotContains(t, rec.Body.String(), "sensitive mnemonic")
}
```

This test currently fails because the request reaches the wallet handler and returns the wallet without authentication.

wolf31o2:
Merged current main into the branch to resolve conflicts after the recent Bursa merges. No source changes beyond the existing wallet-auth fix.

wolf31o2:
Addressed the current-head request-changes finding. GCP-backed wallet routes now fail closed when JWT/JWKS auth is absent, startup validation requires authentication plus at least one non-empty admin subject (including rejecting a list containing only empty strings), and the handler regression test proves wallet material is not returned unauthenticated. Focused tests pass: `go test ./internal/api -run 'TestValidateAPIExposureRequiresGCPWalletAuthentication|TestGCPWalletRoutesFailClosedWithoutAuth|TestRegisterAPIHandlersProtectsGCPWalletRoutes|TestGCPWalletRoutesAuthorizedOperations' -count=1`. Please re-review commit `b3198be`.

wolf31o2:
Addressed the two current-head CodeRabbit findings in commit `366e9fe`. The request field descriptions now come from the Swagger source and `openapi.sh` regeneration produces complete four-column Markdown tables for SignData, TxAssemble, and WalletRestore docs. The reported script-network enum issue was stale: `openapi/api/openapi.yaml` already contains `mainnet`, `preprod`, and `preview` at the current head. Focused API authorization tests remain green.

wolf31o2:
CI lint identified one formatting issue introduced by the documented field comments. Fixed in `72c5a31`; only the affected struct alignment changed.

arepala-uml (approved):

PR #835, opened 2026-09-05
fix(script): cryptographically verify native-script signatures

### LINKED ISSUE #727: script: verify native-script signatures cryptographically
## Problem

- `ValidateScript` treats any 64-byte blob as satisfying a native signature script when `requireSignatures=true`.
- `validateScriptSig` does not verify Ed25519 signatures.
- The script key hash is not checked.
- `/api/script/validate` can return valid for unrelated signature bytes.

## Acceptance criteria

- Define the signed payload in the validation request.
- Verify each Ed25519 signature against its public key.
- Require the public-key hash to match the native script.
- Reject random bytes, wrong keys, and wrong payloads.
- Keep structural-only validation explicitly separate.

cryptodj413:
## Problem

`ValidateScript`'s `ScriptSig` check only verified that a witness was 64
bytes long. Any 64-byte blob satisfied a script's signature requirement:
no public key, no key-hash check, no actual Ed25519 verification. As a
result, `/api/script/validate` could return `valid: true` for unrelated
signature bytes.

## Fix

- `ValidateScript`/`validateScriptSig` now take a `message []byte` (the
  signed payload) plus `[]ScriptWitness` (vkey + signature pairs, aliased
  to gouroboros's `VkeyWitness`). A `ScriptSig` node only validates when
  some witness's Blake2b-224 key hash matches the script's key hash *and*
  its Ed25519 signature verifies against `message`. Random bytes, a
  witness for an unrelated key, and a signature over the wrong payload are
  all rejected.
- Structural-only validation (`requireSignatures=false`) is unchanged.
- HTTP API: `ScriptValidateRequest` gained `message` and `public_keys`
  (parallel array to `signatures`); `handleScriptValidate` builds
  witnesses from them and rejects mismatched-length key/signature arrays.
- CLI: `script validate` gained `--public-keys`, `--message`/
  `--message-hex` (mutually exclusive), paired positionally with
  `--signatures`.
- Regenerated `docs/swagger.{json,yaml}` and `docs/docs.go`; hand-updated
  the generated `openapi/` client model and its doc page to match (no
  Java/openapi-generator toolchain available in this environment).

## Testing

- `go build ./...`, `go vet ./...`, `go test ./...`, `go test -race ./...`
  — pass (root module and `ui/`, `openapi/` nested modules)
- `golangci-lint run ./...` — 0 issues
- `nilaway -include-pkgs=github.com/blinklabs-io/bursa ./...` — clean
- `go test ./docs/...` — pass (swagger/docs consistency check)
- Added `TestValidateScriptWithSignatures` cases and
  `TestHandleScriptValidate_CryptographicVerification` (API-level)
  covering: valid witness verifies, random signature bytes rejected,
  signature from the wrong key rejected, signature over the wrong payload
  rejected. Confirmed both fail without the fix by temporarily
  reintroducing the old length-only check, then pass again with the fix
  restored.

Fixes #727

---
## Summary by cubic
Fixes #727 by replacing the length-only signature check in `ValidateScript` with real Ed25519 verification. Previously any 64-byte blob satisfied a script's signature requirement; now a witness must match the script's key hash (Blake2b-224) and verify against a supplied message payload.

**API and CLI**
- `ScriptValidateRequest` gains hex `message` and `public_keys` (parallel to `signatures`); the CLI gains `--public-keys`, `--message`, and `--message-hex`.
- Swagger docs and the `openapi` client model were updated to match.
- The CLI exits 1 with a diagnostic when a signature-requiring script runs with zero witnesses, including nested multisig leaves; this check runs after validation so witness-free valid branches are unaffected.

**Breaking changes**
- `ValidateScript` now takes `message []byte` and `[]ScriptWitness` instead of `[][]byte`.
- `public_keys` and `signatures` must have equal length; `message` is required when signatures are present and `require_signatures` is true.
- Scripts with signature requirements now fail unless a matching witness verifies; structural-only validation (`requireSignatures=false`) is unchanged.
- The per-node minimum signature counts were removed; a single witness can now satisfy repeated signature leaves.

<sup>Written for commit f5af4bb17275c2dd3211529311101213e5a5a5d6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/bursa/pull/835?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Script validation now verifies Ed25519 signatures against supplied public keys and message payloads.
  - Added message and public-key inputs to API and command-line script validation.
  - Added support for plain-text or hexadecimal message input, with mutually exclusive options.

- **Bug Fixes**
  - Invalid signatures, mismatched keys, and signatures for altered messages are now rejected.

- **Documentation**
  - Updated API schemas and generated documentation for the new validation fields.

### DIFF
--- a/bursa.go
+++ b/bursa.go
@@ -2070,15 +2070,6 @@ func validateScriptAllWithDepth(
 	requireSignatures bool,
 	depth int,
 ) bool {
-	if requireSignatures {
-		total := 0
-		for _, subScript := range script.Scripts {
-			total += minSignaturesRequired(&subScript)
-		}
-		if len(witnesses) < total {
-			return false
-		}
-	}
 	for _, subScript := range script.Scripts {
 		if !validateScriptWithDepth(
 			&subScript,
@@ -2130,22 +2121,6 @@ func validateScriptNOfWithDepth(
 	// Safety check: N should be reasonable for cryptographic purposes
 	if script.N > 255 {
 		return false
-	}
-	if requireSignatures {
-		// Lower bound on signatures needed for this node (accounts for 0-sig leaves like timelocks)
-		// Create a temporary NativeScript to call minSignaturesRequired
-		cborData, err := cbor.Encode(script)
-		if err != nil {
-			return false
-		}
-		var tempScript NativeScript
-		if _, err := cbor.Decode(cborData, &tempScript); err != nil {
-			return false
-		}
-		minReq := minSignaturesRequired(&tempScript)
-		if len(witnesses) < minReq {
-			return false
-		}
 	}
 	satisfied := 0
 	for _, subScript := range script.Scripts {
--- a/bursa.go
+++ b/bursa.go
@@ -95,6 +95,13 @@ type (
 	NativeScriptInvalidHereafter = lcommon.NativeScriptInvalidHereafter
 )
 
+// ScriptWitness pairs an Ed25519 verification key with its signature over the
+// script validation payload, mirroring a transaction's vkey witness. Both
+// fields are required for cryptographic validation: the key hash (Blake2b-224
+// of Vkey) must match a NativeScriptPubkey's key hash, and Signature must
+// verify against Vkey and the supplied message.
+type ScriptWitness = lcommon.VkeyWitness
+
 // GetScriptType returns the script type identifier for a native script.
 // Only accepts *NativeScript types; returns an error for other script types.
 func GetScriptType(script Script) (int, error) {
@@ -1910,18 +1917,23 @@ func GetScriptAddress(script Script, networkName string) (string, error) {
 // Maximum recursion depth for script validation to prevent stack overflow
 const maxScriptDepth = 20
 
-// ValidateScript checks if a script is satisfied given signatures and current slot
-// If requireSignatures is true, requires signatures for ScriptSig scripts and validates format.
-// If false, allows empty signatures for structural validation only.
+// ValidateScript checks if a script is satisfied given witnesses and current slot.
+// message is the signed payload (e.g. a transaction body hash) that each
+// witness's signature is verified against. If requireSignatures is true,
+// ScriptSig scripts require a witness whose verification key hash matches the
+// script and whose signature verifies against message. If false, allows empty
+// witnesses for structural validation only.
 func ValidateScript(
 	script Script,
-	signatures [][]byte,
+	message []byte,
+	witnesses []ScriptWitness,
 	slot uint64,
 	requireSignatures bool,
 ) bool {
 	return validateScriptWithDepth(
 		script,
-		signatures,
+		message,
+		witnesses,
 		slot,
 		requireSignatures,
 		0,
@@ -1977,7 +1989,8 @@ func minSignaturesRequired(script *NativeScript) int {
 // validateScriptWithDepth is the internal recursive validator with depth tracking
 func validateScriptWithDepth(
 	script Script,
-	signatures [][]byte,
+	message []byte,
+	witnesses []ScriptWitness,
 	slot uint64,
 	requireSignatures bool,
 	depth int,
@@ -1994,13 +2007,13 @@ func validateScriptWithDepth(
 
 	switch s := nativeScript.Item().(type) {
 	case *NativeScriptPubkey:
-		return validateScriptSig(s, signatures, requireSignatures)
+		return validateScriptSig(s, message, witnesses, requireSignatures)
 	case *NativeScriptAll:
-		return validateScriptAllWithDepth(s, signatures, slot, requireSignatures, depth+1)
+		return validateScriptAllWithDepth(s, message, witnesses, slot, requireSignatures, depth+1)
 	case *NativeScriptAny:
-		return validateScriptAnyWithDepth(s, signatures, slot, requireSignatures, depth+1)
+		return validateScriptAnyWithDepth(s, message, witnesses, slot, requireSignatures, depth+1)
 	case *NativeScriptNofK:
-		return validateScriptNOfWithDepth(s, signatures, slot, requireSignatures, depth+1)
+		return validateScriptNOfWithDepth(s, message, witnesses, slot, requireSignatures, depth+1)
 	case *NativeScriptInvalidBefore:
 		return validateScriptInvalidBefore(s, slot)
 	case *NativeScriptInvalidHereafter:
@@ -2010,43 +2023,49 @@ func validateScriptWithDepth(
 	}
 }
 
-// validateScriptSig checks if a signature script is satisfied
+// validateScriptSig checks if a signature script is satisfied.
 //
-// SECURITY WARNING: This function currently allows empty signatures for structural validation.
-// In production spending validation, signatures MUST be provided and cryptographically verified.
-// This implementation is a placeholder for testing script logic and should not be used
-// for actual transaction validation without proper Ed25519 signature verification.
+// If requireSignatures is true, at least one witness must carry a
+// verification key whose Blake2b-224 hash matches the script's key hash and
+// whose Ed25519 signature verifies against message. Random bytes, a witness
+// for an unrelated key, or a signature over the wrong payload are all
+// rejected because the cryptographic check fails.
 //
-// TODO: Implement full Ed25519 signature verification against transaction hash and script.Hash
+// If requireSignatures is false, only the script's structure is considered
+// (structural-only validation); no witness is inspected.
 func validateScriptSig(
-	_ *NativeScriptPubkey,
-	signatures [][]byte,
+	script *NativeScriptPubkey,
+	message []byte,
+	witnesses []ScriptWitness,
 	requireSignatures bool,
 ) bool {
-	if requireSignatures {
-		// Require signatures for format validation
-		if len(signatures) == 0 {
-			return false
-		}
-		// TODO: Implement proper Ed25519 signature verification against script.Hash
-		// For now, check basic signature format (64 bytes for Ed25519)
-		for _, sig := range signatures {
-			if len(sig) != 64 {
-				return false // Invalid Ed25519 signature length
-			}
-		}
-		return len(signatures) >= 1
-	} else {
-		// For structural validation, ignore all signature checks
-		// Only validate the script structure itself
+	if !requireSignatures {
+		// Structural validation only: ignore witnesses entirely.
 		return true
 	}
+	if len(script.Hash) != 28 {
+		return false
+	}
+	for _, witness := range witnesses {
+		if len(witness.Vkey) != ed25519.PublicKeySize {
+			continue
+		}
+		keyHash := lcommon.Blake2b224Hash(witness.Vkey)
+		if !bytes.Equal(keyHash.Bytes(), script.Hash) {
+			continue
+		}
+		if lcommon.VerifyVKeySignature(witness.Vkey, witness.Signature, message) == nil {
+			return true
+		}
+	}
+	return false
 }
 
 // validateScriptAllWithDepth checks if all sub-scripts are satisfied
 func validateScriptAllWithDepth(
 	script *NativeScriptAll,
-	signatures [][]byte,
+	message []byte,
+	witnesses []ScriptWitness,
 	slot uint64,
 	requireSignatures bool,
 	depth int,
@@ -2056,14 +2075,15 @@ func validateScriptAllWithDepth(
 		for _, subScript := range script.Scripts {
 			total += minSignaturesRequired(&subScript)
 		}
-		if len(signatures) < total {
+		if len(witnesses) < total {
 			return false
 		}
 	}
 	for _, subScript := range script.Scripts {
 		if !validateScriptWithDepth(
 			&subScript,
-			signatures,
+			message,
+			witnesses,
 			slot,
 			requireSignatures,
 			depth,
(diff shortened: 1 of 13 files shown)


bursa.go
```diff
@@ -2056,14 +2075,15 @@ func validateScriptAllWithDepth(
 		for _, subScript := range script.Scripts {
 			total += minSignaturesRequired(&subScript)
 		}
-		if len(signatures) < total {
+		if len(witnesses) < total {
```
wolf31o2: `minSignaturesRequired` counts signature leaves, not distinct keys, so `len(witnesses)` is not a valid lower bound. Verified on this branch: `all [sig(k), sig(k)]` with the single witness that satisfies both leaves returns `false`, and the same holds for the `NofK` prefilter at line 2146 with `2-of-2 [sig(k), sig(k)]`. The ledger evaluates each `RequireSignature` leaf independently against the witness key-hash set, so both scripts are satisfied by one witness.

The prefilter also blocks nothing: appending an unrelated witness (32-byte vkey, 64-byte signature) to that same input makes both scripts return `true`, because the count passes and the per-leaf crypto check then succeeds on the real witness. It rejects only satisfiable inputs.

Remove both count prefilters and rely on the per-leaf cryptographic check.

The list-of-blobs shape made this bound approximately right before; a witness set makes it wrong, so the change to `[]ScriptWitness` is what exposes it.

internal/api/api.go
```diff
@@ -1576,33 +1585,59 @@ func handleScriptValidate(w http.ResponseWriter, r *http.Request) {
...
-					err,
-				)
-				w.Header().Set("Content-Type", "application/json")
-				w.WriteHeader(http.StatusBadRequest)
-				_, _ = w.Write([]byte(`{"error":"Invalid signature format"}`))
-				return
-			}
-			signatures[i] = sig
+	// Public keys and signatures are parallel arrays forming vkey witnesses.
+	if len(req.PublicKeys) != len(req.Signatures) {
```
wolf31o2: This runs before `req.RequireSignatures` is consulted, so a request carrying `signatures` without `public_keys` now returns 400 even in structural mode, where witnesses are ignored entirely. That request previously returned 200 with a structural result. The PR description states structural-only validation is unchanged; this is a behavioral break for existing clients that is not required by the security fix.

Gate the pairing check and the witness construction loop on `req.RequireSignatures`.

wolf31o2 (changes requested):
The signature verification itself is correct: `validateScriptSig` matches Blake2b-224 of the witness vkey against `NativeScriptPubkey.Hash` and calls `lcommon.VerifyVKeySignature` (real `ed25519.Verify`) against `message`, and every combinator recurses through that same path. Confirmed locally that `TestValidateScriptWithSignatures` and `TestHandleScriptValidate_CryptographicVerification` fail when the length-only check is restored. Two defects below.

bursa.go
```diff
@@ -2010,60 +2023,58 @@ func validateScriptWithDepth(
...
-			return false
-		}
-	}
 	for _, subScript := range script.Scripts {
 		if !validateScriptWithDepth(
 			&subScript,
-			signatures,
+			message,
+			witnesses,
 			slot,
```
wolf31o2, line 2031: Blocking: `minSignaturesRequired` counts signature leaves, not distinct witnesses. A single matching witness can satisfy repeated-key leaves in `all [sig(k), sig(k)]` or a threshold branch, but this count prefilter rejects those valid scripts before recursive cryptographic validation. Remove this prefilter and the corresponding N-of-K count check.

wolf31o2 (changes requested):

bursa.go
```diff
@@ -2056,14 +2075,15 @@ func validateScriptAllWithDepth(
 		for _, subScript := range script.Scripts {
 			total += minSignaturesRequired(&subScript)
 		}
-		if len(signatures) < total {
+		if len(witnesses) < total {
 			return false
```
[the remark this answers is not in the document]
cryptodj413: Fixed in 9cf8ec2 (removed both count prefilters, trusting the per-leaf recursive/cryptographic check) and c8405b4 (deleted the now-dead `minSignaturesRequired` helper and its `sort` import, which golangci-lint's `unused` check flagged once its only callers were gone). Added `TestValidateScriptWithSignaturesRepeatedKey`-equivalent coverage (repeated-key `all`/`NofK` assertions in `TestValidateScriptWithSignatures`) confirming a single witness satisfies same-key leaves.

docs/docs.go
```diff
@@ -1153,6 +1153,15 @@ const docTemplate = `{
...
             "properties": {
+                "message": {
+                    "type": "string"
+                },
+                "public_keys": {
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    }
+                },
```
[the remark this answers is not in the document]
cryptodj413: Fixed in 9cf8ec2: `ScriptValidateRequest.Message`/`PublicKeys` now carry `format:"hex"` and `minLength"64" maxLength:"64"` struct tags, regenerated into `docs/swagger.{json,yaml}`/`docs.go` (public_keys items now declare `format: hex`, `minLength: 64`, `maxLength: 64`). `Signatures` keeps its pre-existing `validate:"dive,hexadecimal,len=128"` tag but wasn't part of this PR's schema change.

bursa.go
cryptodj413: Agreed and fixed — this matches your 9cf8ec2 commit exactly (removed both count prefilters). Follow-up in c8405b4: deleted the now-dead `minSignaturesRequired` and its `sort" import, since golangci-lint's `unused` check flagged it once both callers were gone (confirmed clean: `golangci-lint run ./...` → 0 issues). Current head: c8405b4.

internal/api/api.go
cryptodj413: Agreed and fixed — matches your 9cf8ec2 commit (public_keys/signatures pairing check and witness-building loop are now gated on req.RequireSignatures, so a structural-only request carrying only the legacy signatures field is not rejected). Added TestHandleScriptValidate_StructuralOnlyIgnoresLegacySignatures in c8405b4 covering exactly this case. Current head: c8405b4.

bursa.go
cryptodj413: Fixed as of c8405b4: minSignaturesRequired and both count prefilters are gone (your 9cf8ec2 removed the prefilters; c8405b4 removes the now-dead helper itself plus the sort import, since golangci-lint's unused check flagged it once both callers were gone). Verified: golangci-lint run ./... -> 0 issues, go test -race ./... passes, and TestValidateScriptWithSignatures now asserts the repeated-key all/NofK cases validate with a single witness.

internal/cli/cli.go
```diff
@@ -1874,26 +1886,35 @@ func RunScriptValidate(
...
+		witnesses = append(
+			witnesses,
+			bursa.ScriptWitness{Vkey: vkey, Signature: sig},
+		)
 	}
 
 	// Check if signatures are required
-	if !structuralOnly && len(sigBytes) == 0 &&
-		scriptRequiresSignatures(script) {
+	if !structuralOnly && len(witnesses) > 0 && len(messageBytes) == 0 &&
```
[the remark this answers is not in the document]
cryptodj413: Evaluated, declining: this is a UX/diagnostics observation, not a correctness or security regression. ValidateScript still returns the correct, safe result (valid:false) for a signature-requiring script run with no witnesses — the exit-0 change is intentional, not accidental.

Restoring a script-aware preflight (per the suggestion, accounting for nested any/timelock semantics) means re-implementing ValidateScript's own slot/threshold/key-hash evaluation a second time, just to decide whether to print an error before calling it. That's the same class of bug this PR fixes: a heuristic that can diverge from the real evaluator and reject (or here, mis-diagnose) a script the real evaluator would handle correctly. wolf31o2's 9cf8ec2 removed the equivalent preflight for exactly this reason (see the CodeRabbit/wolf31o2 threads on bursa.go), and I removed its bursa.go counterpart the same way — keeping the CLI's removal consistent with that decision rather than reintroducing an approximation on one side only.

Net effect: the CLI is fail-secure (never wrongly validates something invalid), just without a distinct "you forgot --message/witnesses" message before the fact. Happy to revisit if wolf31o2 wants the diagnostic back for a narrower, provably-safe subset (e.g. only when the outer script type is a bare sig/all with no any/timelock anywhere in the tree).

cryptodj413:
Addressed all open review threads as of c8405b4:

- Removed the two witness-count prefilters (bursa.go) per wolf31o2/CodeRabbit; deleted the now-dead `minSignaturesRequired` helper golangci-lint flagged as unused once both callers were gone.
- Gated the public_keys/signatures pairing and parsing in `handleScriptValidate` behind `RequireSignatures` per wolf31o2, so structural-only requests using the legacy `signatures`-only shape aren't rejected.
- Documented the message/public_keys hex encoding contract in the generated swagger docs (already done in 9cf8ec2).
- Declined cubic's suggestion to restore a script-aware CLI preflight (rationale on that thread): the current behavior is fail-secure, and a heuristic preflight is the same class of bug this PR fixes.
- Added regression coverage: repeated-key `all`/`NofK` cases in `TestValidateScriptWithSignatures`, and `TestHandleScriptValidate_StructuralOnlyIgnoresLegacySignatures`.

Validated on c8405b4: `go build/vet/test -race ./...`, `golangci-lint run ./...` (0 issues), `nilaway` (clean), all pass.

internal/cli/cli.go
wolf31o2, line 1942: When default cryptographic validation is requested and the script contains a signature leaf, `len(witnesses) == 0` makes this guard skip and the command proceeds to print `valid:false` with exit status 0. That contradicts the command's documented contract that witnesses are required by default and regresses the previous CLI behavior for missing signatures. Please restore a script-aware preflight error for an unsatisfied signature-only path, while retaining the valid witness-free path for a conditional script whose timelock branch is satisfied.

wolf31o2 (changes requested):

internal/cli/cli.go
cryptodj413: Fixed in 8c8e018. Restored the diagnostic, but sequenced after ValidateScript rather than before it: call ValidateScript first (real, slot-aware evaluator), and only when it has already returned false with zero witnesses AND the script contains a signature leaf, exit 1 with the restored error instead of printing the JSON result. Since the check now runs after the real evaluator, it can only refine the message for an already-correct false — it can't override a true from a witness-free branch (e.g. any[sig, before(faraway)]).

Verified manually with the built CLI binary:
- plain sig script, no witnesses, --slot 1000 -> exits 1 with "signatures required for format validation..."
- any[sig, before(999999999)], no witnesses, --slot 1000 -> exits 0, {"valid": true} (before-branch satisfied)
- same sig-only script with --structural-only -> exits 0, {"valid": true} (unaffected)

Re-added scriptRequiresSignatures/anyScriptRequiresSignatures/convertScripts with a doc comment stating they're a diagnostic-message selector only, never a gate on validity. Current head: 8c8e018.

cryptodj413:
Addressed wolf31o2's latest review (8c8e018): restored the CLI's signature-required preflight error, but sequenced *after* `ValidateScript` runs rather than before — so it can only add a clearer diagnostic to an already-correct `false`, never override a script that validates `true` via a witness-free branch. Manually verified all three cases (plain sig script with no witnesses now hard-errors again; a witness-free conditional script with a satisfied timelock still validates; `--structural-only` unaffected) with the built CLI binary — details on the thread.

Full validation on 8c8e018: `go build/vet/test -race ./...`, `golangci-lint run ./...` (0 issues) all pass.

internal/cli/cli.go
```diff
@@ -1779,6 +1779,12 @@ func RunScriptCreate(
 	return nil
```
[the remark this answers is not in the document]
cryptodj413: Confirmed and fixed in f5af4bb. convertScripts now takes &scripts[i] instead of copying scr into the interface, so scriptRequiresSignatures' `*bursa.NativeScript` type assertion succeeds for nested nodes.

Reproduced the exact bug first: a 2-of-3 bursa.NewMultiSigScript run with no witnesses silently printed valid:false at exit 0 instead of the restored diagnostic, because every nested sig leaf inside the NofK wrapper failed the pointer type assertion. After the fix, the same script correctly exits 1 with the "signatures required..." error; supplying any witness (even a non-satisfying one) still falls through to the normal valid:false/true path unaffected; the witness-free timelock-branch case from 8c8e018 is unaffected.

Added TestScriptRequiresSignatures_Nested (asserts a nested nOf and a nested all both report true, a pure timelock reports false), confirmed it fails against the value-copy version and passes with the pointer fix. Current head: f5af4bb.

cryptodj413:
Addressed cubic's new finding (f5af4bb): `convertScripts` was storing value-typed `NativeScript` nodes instead of pointers, so `scriptRequiresSignatures`'s `*bursa.NativeScript` type assertion silently failed for every nested script — meaning the witness-required preflight from 8c8e018 never actually fired for an ordinary multisig (nested nOf/all), only for a bare unwrapped single-sig script. Fixed by taking the address of each slice element. Reproduced the bug first (built the CLI, ran a real 2-of-3 multisig with no witnesses, confirmed it silently returned `valid:false` at exit 0), then confirmed the fix restores exit 1 with the diagnostic, and that supplying any witness / the witness-free-timelock case are both unaffected. Added `TestScriptRequiresSignatures_Nested`, confirmed it fails without the fix.

Full validation on f5af4bb: `go build/vet/test -race ./...`, `golangci-lint run ./...` (0 issues) all pass.

wolf31o2 (approved):

wolf31o2 (approved):

PR #839, opened 2026-09-07
fix(poolops): reject KES issue counter overflow

wolf31o2:
## Summary

- Reject KES issue-number overflow before rotating an operational certificate.
- Return the existing invalid-request error and cover the boundary with a regression test.

## Validation

- Focused poolops tests and race tests passed.
- Root `go test -race ./...` passed.
- `git diff --check` passed.

This is a bounded DB-ACT-13 fix. The report’s complete key-type, official-vector, downstream-consumer, and release-chain validation remains open.

---
## Summary by cubic
Rejects KES issue counter overflow, malformed VRF/KES export material, invalid pool certificate inputs, and unsafe key files so counters can't wrap to zero, malformed keys fail early, bad margins return errors instead of panicking, and secret keys aren't loaded from non-regular or permissive files.

- Returns `ErrInvalidRequest` when the previous issue number is `math.MaxUint64`.
- Validates VRF and KES key sizes and KES depth in export helpers.
- Validates pool registration and retirement certificate inputs: nil pointers and margin bounds.
- Hardens key file loading: rejects symlinks and reparse points, caps file size, and enforces restrictive permissions on secret keys.
- Adds regression tests for overflow, invalid VRF/KES material, invalid margins, and file loading edge cases, plus an authoritative envelope vector suite covering all exported key types with trailing-CBOR, wrong-length, and mismatched-identity rejection.

<sup>Written for commit 19bdf28f1ae513f499fe9d10b6da1f435d11903e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/bursa/pull/839?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/key_file_test.go
+++ b/key_file_test.go
@@ -15,7 +15,6 @@
 package bursa
 
 import (
-	"encoding/json"
 	"os"
 	"path/filepath"
 	"strings"
@@ -49,17 +48,4 @@ func TestLoadKeyFromFileRejectsOversizedInput(t *testing.T) {
 	assert.Nil(t, key)
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "exceeds maximum size")
-}
-
-func TestLoadKeyFromFileRejectsPermissiveSecret(t *testing.T) {
-	wallet, err := NewWallet(testSecretKeyMnemonic)
-	require.NoError(t, err)
-	data, err := json.Marshal(wallet.PaymentSKey)
-	require.NoError(t, err)
-	path := filepath.Join(t.TempDir(), "payment.skey")
-	require.NoError(t, os.WriteFile(path, data, 0o644))
-
-	key, err := LoadKeyFromFile(path)
-	assert.Nil(t, key)
-	assert.ErrorIs(t, err, ErrInsecureFileMode)
 }
--- a/key_file_test.go
+++ b/key_file_test.go
@@ -1,0 +1,65 @@
+// Copyright 2026 Blink Labs Software
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package bursa
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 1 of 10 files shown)


ui/internal/poolops/service.go
```diff
@@ -288,6 +289,9 @@ func (s *Service) IssueOpCert(password string, kesIndex uint32, issueNumber, kes
 // counter incremented. The current issue number is supplied by the caller (read
 // from the previous opcert / counter file); the new cert uses prevIssue+1.
 func (s *Service) RotateKES(password string, newKESIndex uint32, prevIssueNumber, kesPeriod uint64) (OpCert, error) {
+	if prevIssueNumber == math.MaxUint64 {
```
arepala-uml, line 292: Verified: the overflow check runs before the `+1` increment, so `prevIssueNumber` can't already have wrapped when this fires. `math.MaxUint64` matches the field's real width all the way through `IssueOpCert` -> `CreateOperationalCertificate` -> `ledger.CreateOpCert` (uint64 throughout, no narrowing cast), so this is the true boundary, not an arbitrarily-large stand-in. The regression test drives the counter to that exact value. Root-cause fix, correctly placed.

key_file_test.go
```diff
@@ -0,0 +1,65 @@
...
+	wallet, err := NewWallet(testSecretKeyMnemonic)
+	require.NoError(t, err)
+	data, err := json.Marshal(wallet.PaymentSKey)
+	require.NoError(t, err)
+	path := filepath.Join(t.TempDir(), "payment.skey")
+	require.NoError(t, os.WriteFile(path, data, 0o644))
+
+	key, err := LoadKeyFromFile(path)
+	assert.Nil(t, key)
+	assert.ErrorIs(t, err, ErrInsecureFileMode)
```
arepala-uml: Non-blocking, re: cubic's finding on this line. Confirmed valid by reading the platform split: `key_permissions_unix.go`/`key_permissions_windows.go` implement `checkOpenFilePermissions` very differently (POSIX mode bits vs. DACL), and this new file carries no `//go:build unix` tag, unlike the existing `key_file_permissions_unix_test.go` (which does tag `unix`). So `TestLoadKeyFromFileRejectsPermissiveSecret` asserts a Unix-only outcome (`0o644` -> `ErrInsecureFileMode`) that doesn't hold under Windows' DACL-inheritance-from-tempdir behavior.

This can't fail CI today: `go-test.yml`'s matrix is `platform: [ubuntu-latest]` only, and the `windows` entries that do exist (cross-build, publish's build-msi) only `go build`, never `go test`. So it's a latent portability/flakiness issue, not a merge blocker. Recommend moving this one assertion under the existing unix tag (or into `key_file_permissions_unix_test.go` alongside `TestReadSecretKeyFileRejectsFIFO`) before Windows `go test` coverage is ever added.

arepala-uml (dismissed):

arepala-uml (approved):

PR #849, opened 2026-09-07
fix(signer): bound request authentication reads

wolf31o2:
## Summary

- Add a request read deadline for signer authentication bodies.
- Reject malformed signature inputs before reading the request body.
- Preserve body hashing, timestamp, signature, and replay checks.

## Validation

- Focused signer race tests passed repeatedly.
- Root race suite and API vet passed.

Live deployed-process behavior remains outside this change.

---
## Summary by cubic
Fixes the signer's request authentication so slow clients can't hold body reads open indefinitely and malformed signatures are rejected before the body is read.

- Adds a 30-second read timeout to the signer HTTP server.
- Validates signature length and hex encoding before buffering the request body.
- Preserves body hashing, timestamp, signature, and replay checks.

<sup>Written for commit 28f6c7fd9dbf66d16a0106f80e79c2ab03442756. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/bursa/pull/849?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/cmd/bursa/signer_test.go
+++ b/cmd/bursa/signer_test.go
@@ -15,10 +15,21 @@
 package main
 
 import (
+	"net/http"
 	"testing"
 
 	"github.com/blinklabs-io/bursa/internal/signer"
 )
+
+func TestNewSignerHTTPServerReadTimeout(t *testing.T) {
+	srv := newSignerHTTPServer("127.0.0.1:0", http.NotFoundHandler())
+	if srv.ReadHeaderTimeout <= 0 {
+		t.Fatal("signer server must bound header reads")
+	}
+	if srv.ReadTimeout != signerReadTimeout {
+		t.Fatalf("ReadTimeout = %s, want %s", srv.ReadTimeout, signerReadTimeout)
+	}
+}
 
 func TestIsLocalListenAddress(t *testing.T) {
 	tests := []struct {
--- a/cmd/bursa/signer.go
+++ b/cmd/bursa/signer.go
@@ -39,6 +39,20 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const signerReadTimeout = 30 * time.Second
+
+func newSignerHTTPServer(addr string, handler http.Handler) *http.Server {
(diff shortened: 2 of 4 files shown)


cmd/bursa/signer_test.go
```diff
@@ -15,11 +15,22 @@
...
 	"github.com/blinklabs-io/bursa/internal/signer"
 )
 
+func TestNewSignerHTTPServerReadTimeout(t *testing.T) {
+	srv := newSignerHTTPServer("127.0.0.1:0", http.NotFoundHandler())
+	if srv.ReadHeaderTimeout <= 0 {
+		t.Fatal("signer server must bound header reads")
+	}
+	if srv.ReadTimeout != signerReadTimeout {
+		t.Fatalf("ReadTimeout = %s, want %s", srv.ReadTimeout, signerReadTimeout)
```
arepala-uml, line 30: Non-blocking: this only asserts the `http.Server.ReadTimeout` field equals `signerReadTimeout` — it doesn't exercise the bound behaviorally (e.g. a slow-trickle client actually getting disconnected). Given `net/http`'s well-established `ReadTimeout` semantics and the cost of a live timing test, the config-level assertion is reasonable here; flagging only in case a follow-up wants a behavioral test (e.g. a short injected timeout against a real listener) for extra confidence.

arepala-uml (approved):

cmd/bursa/signer_test.go
wolf31o2: The reviewer’s note is non-blocking. The test verifies the configured ReadTimeout, whose enforcement is provided by net/http; no behavioral timing test is needed for this change.

PR #854, opened 2026-09-07
fix(signer): bound PKCS#11 signing waits

wolf31o2:
## Summary

- Forward cancellation through PKCS#11 signing.
- Serialize native session use through a single-owner worker.
- Bound queued signing work and prevent teardown races.

## Validation

- Cancellation and concurrency regressions pass.
- Tagged backend race suites, SoftHSM signing, ambiguity tests, and vet pass.

A synchronous native `C_Sign` call can still delay worker teardown; real hardware and non-Linux token behavior remain outside this change.

---
## Summary by cubic
Bounded PKCS#11 signing waits so a wedged token can no longer hang signing indefinitely, and serialized native session use through a single worker so concurrent calls can't corrupt the session. When the caller passes an unbounded context, the backend now applies its own 30s ceiling.

**Details**
- Requests queue behind one worker; callers can abort waiting on a full queue, and canceled queued requests are never sent to the token.
- Cancellation is honored while queued and during the native call; an already-started operation still completes.
- `Close` waits for in-flight signing before teardown.
- Added a CI job that runs the tagged PKCS#11 tests with `-race`.
- A synchronous native `C_Sign` call can still delay worker teardown; real hardware and non-Linux token behavior remain outside this change.

<sup>Written for commit df5899d20e7b6502c19180b991ad38eb6f840868. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/bursa/pull/854?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * PKCS#11 signing operations now support caller cancellation, allowing applications to stop active or queued requests promptly.
  * Signing requests are processed in an orderly sequence to improve compatibility with PKCS#11 tokens.

* **Bug Fixes**
  * Prevented canceled queued requests from being sent to the token.
  * Improved backend shutdown handling so in-progress work completes cleanly before resources are released.

### DIFF
--- a/internal/signer/backend/pkcs11.go
+++ b/internal/signer/backend/pkcs11.go
@@ -50,7 +50,7 @@ type pkcs11Key struct {
 	typ         KeyType
 	backendName string
 	priv        pkcs11.ObjectHandle
-	sign        func(priv pkcs11.ObjectHandle, msg []byte) ([]byte, error)
+	sign        func(ctx context.Context, priv pkcs11.ObjectHandle, msg []byte) ([]byte, error)
 }
 
 func (k *pkcs11Key) Hash() KeyHash                { return k.hash }
@@ -63,8 +63,11 @@ func (k *pkcs11Key) Backend() string              { return k.backendName }
 // CKM_EDDSA with no parameter is PureEdDSA, so the token signs digest as the
 // message exactly like ed25519.Sign — byte-for-byte identical to the software
 // and Vault backends for the same key and message (Ed25519 is deterministic).
-func (k *pkcs11Key) Sign(_ context.Context, digest []byte) ([]byte, error) {
-	sig, err := k.sign(k.priv, digest)
+func (k *pkcs11Key) Sign(ctx context.Context, digest []byte) ([]byte, error) {
+	if ctx == nil {
+		return nil, errors.New("pkcs11 sign: nil context")
+	}
+	sig, err := k.sign(ctx, k.priv, digest)
 	if err != nil {
 		return nil, fmt.Errorf("pkcs11 sign: %w", err)
 	}
@@ -79,10 +82,31 @@ type PKCS11Backend struct {
 	name    string
 	ctx     *pkcs11.Ctx
 	session pkcs11.SessionHandle
-	// mu serializes token operations: a PKCS#11 session is single-threaded and
-	// SignInit/Sign form one non-reentrant operation.
-	mu   sync.Mutex
-	keys map[KeyHash]*pkcs11Key
+	// mu protects closed and the request-channel close. The signing worker is
+	// the sole owner of session and ctx during an operation.
+	mu         sync.Mutex
+	closed     bool
+	requests   chan pkcs11SignRequest
+	workerWG   sync.WaitGroup
+	closeOnce  sync.Once
+	closeDone  chan struct{}
+	closeErr   error
+	nativeSign func(priv pkcs11.ObjectHandle, msg []byte) ([]byte, error)
+	keys       map[KeyHash]*pkcs11Key
+}
+
+const maxQueuedPKCS11Signs = 1
+
+type pkcs11SignRequest struct {
+	ctx  context.Context
+	priv pkcs11.ObjectHandle
+	msg  []byte
+	resp chan pkcs11SignResult
+}
+
+type pkcs11SignResult struct {
+	sig []byte
+	err error
 }
 
 // NewPKCS11Backend loads the module, selects the token, logs in with the PIN,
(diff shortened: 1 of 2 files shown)


internal/signer/backend/pkcs11_worker_test.go
```diff
@@ -0,0 +1,151 @@
...
+		t.Fatal("native signing did not start")
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	queuedDone := make(chan error, 1)
+	go func() {
+		_, err := key.Sign(ctx, []byte("queued"))
+		queuedDone <- err
+	}()
+	cancel()
```
arepala-uml, line 139: Confirmed (also flagged independently by CodeRabbit and Cubic on this head): `cancel()` here has no synchronization guaranteeing the "queued" `key.Sign` goroutine has actually reached the `b.requests <- req` send inside `signWithSession` before cancellation fires. Since `b.requests` (cap 1) is empty at this point — the first request already left the channel into the worker — the send is non-blocking, so depending on scheduling this can go either way:
- enqueued, then dropped by `signWorker`'s `req.ctx.Err()` check (the path this test is named for), or
- never enqueued at all, returning `context.Canceled` from the *first* select in `signWithSession` before `b.requests` is ever touched.

`calls == 1` holds either way, so the assertion doesn't distinguish the two, and the worker's "skip an already-canceled queued request" branch (`pkcs11.go` ~line 217) isn't reliably exercised by this test. Non-blocking for this PR — the sibling test (`TestPKCS11Sign_ContextCancellationDoesNotConcurrentSessionUse`) does correctly synchronize on `<-started` before canceling and solidly proves the core "bound the wait" claim — but worth tightening this one so a regression in the queued-skip path can't slip through on a lucky `select`.

internal/signer/backend/pkcs11_worker_test.go
```diff
@@ -0,0 +1,151 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build pkcs11
+
+package backend
```
arepala-uml, line 17: Nice, deterministic tests for the queue/cancellation behavior (verified: `go test -tags pkcs11 -race ./internal/signer/backend/...` passes, and a manual repro confirms `Sign` returns in ~50ms via `context.DeadlineExceeded` even when the injected native call never returns).

One gap outside this diff: `.github/workflows/go-test.yml` runs `go test ./...` with no `-tags pkcs11`, so this file (and the rest of the CGO-gated PKCS#11 backend, including `pkcs11_softhsm_test.go`) never builds or runs in CI — pre-existing, not introduced here, but worth a follow-up issue since these are exactly the tests that prove the DoS fix works.

internal/signer/backend/pkcs11.go
```diff
@@ -136,13 +160,64 @@ func NewPKCS11Backend(cfg PKCS11Config) (Backend, error) {
...
+	go b.signWorker()
 	return b, nil
 }
 
-// signWithSession runs one SignInit+Sign under the session mutex.
-func (b *PKCS11Backend) signWithSession(priv pkcs11.ObjectHandle, msg []byte) ([]byte, error) {
+// signWithSession queues one signing operation. The caller may stop waiting
+// when ctx is canceled; the worker still completes an already-started native
+// operation before it accepts another request or teardown runs.
+func (b *PKCS11Backend) signWithSession(ctx context.Context, priv pkcs11.ObjectHandle, msg []byte) ([]byte, error) {
```
arepala-uml, line 173: The bound here is real (confirmed: cancellation while queued and while a native op is executing both return promptly, verified with `-race`), but it's entirely dependent on the caller supplying a `ctx` with a deadline/cancellation. Today the only production call site, `handleSign` in `internal/signer/api/server.go` (via `Coordinator.SignTx`), passes `r.Context()` straight through — the signer HTTP server (`cmd/bursa/signer.go`) sets only `ReadHeaderTimeout: 10s`, no request body/handler timeout — so a genuinely wedged HSM would still block a live `/v1/sign` request indefinitely in the deployed server unless the client disconnects.

For reference, the Vault backend already establishes a self-contained pattern for this in the same package: `vault.go:107` wraps the caller's `ctx` with its own `context.WithTimeout(ctx, 30*time.Second)` regardless of what the caller supplies. Worth a follow-up to do the same here (or wire a request-level timeout into the HTTP handler) so this fix's protection actually takes effect for real signing traffic, not just for callers that happen to pass a bounded context. Non-blocking for this PR since it's a net improvement either way and doesn't regress current behavior (previously `ctx` was ignored entirely).

internal/signer/backend/pkcs11.go
```diff
@@ -354,14 +436,27 @@ func (b *PKCS11Backend) ListKeys(_ context.Context) ([]KeyRef, error) {
 
 // Close logs out and finalizes the module, releasing the session.
 func (b *PKCS11Backend) Close() error {
```
arepala-uml, line 438: Confirmed by reproduction: if the native `C_Sign` call never returns (a truly wedged token/session, not just a slow one), `Close()` blocks forever on `b.workerWG.Wait()` — the caller-facing `Sign` call returns promptly on its own context, but the worker goroutine stays parked in the C call and `Close` has no bound at all. This matches what the PR description already discloses ("A synchronous native C_Sign call can still delay worker teardown"), and today `Backend.Close()` isn't part of the `Backend` interface and isn't called from any production path (`BuildBackends` never closes what it constructs), so the blast radius is limited to future callers of `Close()`. Given that, non-blocking — but consider carrying the caveat here as a doc comment (not just the PR description) so a future caller of `Close()` isn't surprised, since the current comment ("Close logs out and finalizes the module, releasing the session") reads as unconditional.

arepala-uml (dismissed):

internal/signer/backend/pkcs11.go
wolf31o2: Addressed in 84ca70e rather than deferred — you are right that the protection did not reach the path that needed it.

signWithSession now imposes its own ceiling when the caller brings no deadline, following vault.go's stance. 30s: an Ed25519 signature is milliseconds of work on a healthy token, so it only ever fires on one that has stopped answering. A caller that does supply a deadline still wins, since the injection is conditional on there being none.

Covered by a test that signs with context.Background() against a native call that never returns, and asserts DeadlineExceeded comes back — it hangs for the full 5s timeout without the change.

internal/signer/backend/pkcs11_worker_test.go
wolf31o2: Fixed here rather than as a follow-up, since it is what makes the rest of this PR verifiable: go-test.yml now has a go-test (pkcs11) job.

Scoped to ./internal/signer/... — the only tagged code — and run with -race, since worker and cancellation concurrency is exactly what these tests cover. No extra packages needed: miekg/pkcs11 links only -ldl on linux.

Verified locally the way CI will run it: go test -tags pkcs11 -race ./internal/signer/... passes across all six signer packages.

internal/signer/backend/pkcs11_worker_test.go
wolf31o2: This one is already addressed on the current head — 9584317 added the wait you and CodeRabbit both asked for: the test now spins until len(b.requests) == 1 before calling cancel, so the queued request is provably in the channel and the path under test is the one signWorker's ctx.Err() check drops.

Verified with go test -tags pkcs11 -race ./internal/signer/backend/ -run PKCS11Sign_Canceled.

internal/signer/backend/pkcs11.go
wolf31o2: Agreed, and leaving it as disclosed — a bound on Close() would have to abandon the worker goroutine mid-C_Sign rather than wait for it, and abandoning a goroutine parked in a C call does not make the token any less wedged; it just moves where the leak is recorded.

Worth noting the blast radius did not change with this PR: Close() is still not part of the Backend interface and BuildBackends never calls it. If it becomes a production path — a reload or a shutdown that must complete — that is the point to give it a deadline and document what is left running, rather than now, when there is no caller to bound.

internal/signer/backend/pkcs11.go
```diff
@@ -136,13 +167,76 @@ func NewPKCS11Backend(cfg PKCS11Config) (Backend, error) {
...
+		msg:  append([]byte(nil), msg...),
+		resp: make(chan pkcs11SignResult, 1),
+	}
 	b.mu.Lock()
-	defer b.mu.Unlock()
+	if b.closed || b.requests == nil {
+		b.mu.Unlock()
+		return nil, errors.New("pkcs11 sign: backend is closed")
+	}
+	select {
```
arepala-uml, line 207: With the 1-slot queue full, a third caller blocks on this send while still holding b.mu (not released until the select resolves), so a fourth concurrent caller reaching only b.mu.Lock() cannot honor its own context deadline until the third caller's turn resolves. Reproduced: a caller with a 30ms deadline returned only after ~480ms, when an unrelated third caller's 500ms context expired instead. Close() can stall the same way. Use a dedicated closing channel in this select instead of holding b.mu across the blocking send, so each caller's own ctx.Done() is evaluated independently.

arepala-uml (changes requested):
Reproduced a concurrency defect: with the sign queue full, holding b.mu across the blocking channel send serializes producers, so a caller behind another in line cannot honor its own context deadline until the one ahead of it resolves -- undermining the PR's own cancellation guarantee under concurrent load.

arepala-uml (approved):

PR #855, opened 2026-09-07
fix(cli): hash metadata bytes and reject invalid margins

wolf31o2:
## Summary

- Hash validated metadata using its exact file/wire bytes.
- Document the distinct metadata hash domains in CLI and UI paths.
- Reject NaN, infinite, negative, and greater-than-one pool margins deterministically.
- Remove the unused root JCS dependency.

## Validation

- Focused CLI race tests and Cardano CLI compatibility tests passed.
- CLI vet, module tidy, and diff checks passed.

Direct lower-level certificate callers and broader release adoption remain separate scope.

---
## Summary by cubic
Changes metadata hashing to use the exact file bytes instead of canonicalized JSON, and tightens pool margin validation. The CLI no longer runs JCS canonicalization before hashing, so formatting changes now change the digest. Pool margins now reject NaN, infinite, negative, and greater-than-one values deterministically. Removes the unused `jcs` dependency.

<sup>Written for commit afc2ea31634b0b461151db4331a44a741c644826. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/bursa/pull/855?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/internal/cli/capability_test.go
+++ b/internal/cli/capability_test.go
@@ -16,6 +16,7 @@ package cli
 
 import (
 	"encoding/json"
+	"math"
 	"os"
 	"path/filepath"
 	"strings"
@@ -159,6 +160,17 @@ func TestRunCertPoolRegistration_InvalidMargin(t *testing.T) {
 	assert.Contains(t, err.Error(), "margin")
 }
 
+func TestRunCertPoolRegistration_RejectsNonFiniteMargins(t *testing.T) {
+	for _, margin := range []float64{math.NaN(), math.Inf(1), math.Inf(-1)} {
+		err := RunCertPoolRegistration(
+			"cold.vkey", "vrf.vkey", goldenStakeAddr, "",
+			1, 1, margin, "", "",
+		)
+		require.Error(t, err, "margin %v", margin)
+		assert.Contains(t, err.Error(), "margin", "margin %v", margin)
+	}
+}
+
 func TestRunCertPoolRegistration_MetadataPairing(t *testing.T) {
 	err := RunCertPoolRegistration(
 		"cold.vkey", "vrf.vkey", goldenStakeAddr, "",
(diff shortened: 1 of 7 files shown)


internal/cli/capability_test.go
```diff
@@ -273,12 +285,26 @@ func TestFloatToRational(t *testing.T) {
...
 
+func TestFloatToRationalRejectsInvalidMargins(t *testing.T) {
+	for _, margin := range []float64{-0.01, 1.01, math.NaN(), math.Inf(1), math.Inf(-1)} {
+		_, _, err := floatToRational(margin)
+		require.Error(t, err, "margin %v", margin)
+		assert.EqualError(
+			t,
+			err,
+			"pool margin must be a finite value between 0.0 and 1.0",
+			"margin %v",
```
arepala-uml, line 303: Non-blocking: this `assert.EqualError` msgAndArgs is just `"margin %v"` with no trailing `margin` argument, so a failure here would print the literal string `margin %v` instead of substituting the actual value (unlike `require.Error` two lines up, which passes `margin` correctly). Doesn't affect correctness, just failure-message clarity.

arepala-uml (approved):

internal/cli/capability_test.go
wolf31o2: The reviewer’s note is non-blocking and limited to failure-message clarity; it does not affect the final behavior or validation covered by this PR.

PR #856, opened 2026-09-07
fix(signer): bound replay cache key storage

wolf31o2:
## Summary

- Store replay identities as fixed-width length-delimited digests.
- Add explicit retained-key byte accounting alongside the entry-count limit.
- Preserve the original nonce in the signed canonical request.

## Validation

- Oversized-nonce regression reproduced before the fix.
- Repeated request-signing/replay race tests and the complete signer API race suite passed.
- Command tests and vet passed.

The retained-cache bound does not change valid signed-request compatibility.

---
## Summary by cubic
Fixes BD3-F3 by bounding the signer replay cache: oversized authenticated nonces are stored as SHA-256 digests, and a retained-key byte budget now applies alongside the entry-count cap. Valid signed-request compatibility is unchanged.

**Bug Fixes**

- Replay identities hash length-delimited caller and nonce fields, so `("a","bc")` and `("ab","c")` stay distinct.
- The cache fails closed when either the entry count or the byte budget is exhausted; the original nonce is still preserved in the signed canonical request.
- Ledger UI test mocks are constructable again.

<sup>Written for commit f05a619a0cd31844529b866bafce67791b21428e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/bursa/pull/856?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved nonce replay protection to prevent collisions between callers and nonces.
  * Added safeguards to limit replay-cache storage by both entry count and memory usage.
  * Improved handling of unusually large nonces and cache capacity limits.
  * Enhanced reliability for concurrent and expired nonce-cache entries.


internal/signer/api/requestsign.go
```diff
@@ -180,53 +191,83 @@ func (a *RequestSigningAuthenticator) readBody(r *http.Request) ([]byte, error)
...
+	mu       sync.Mutex
+	entries  map[[nonceCacheKeyBytes]byte]int64 // digest -> expiry (unix nanos)
+	ttl      time.Duration
+	max      int
+	bytes    int64
+	maxBytes int64
+	now      func() time.Time
 }
 
 func newNonceCache(ttl time.Duration, max int) *nonceCache {
```
arepala-uml, line 207: Merge blocker: the required `lint` check fails on this head. Reproduced locally with an isolated `GOLANGCI_LINT_CACHE`:

```
internal/signer/api/requestsign.go:207:6: func newNonceCache is unused (unused)
```

On `origin/main`, `newNonceCache` has one production call site (`NewRequestSigningAuthenticator` → `cache: newNonceCache(2*skew, defaultNonceCacheMax)`). This PR replaces that call with `newNonceCacheWithByteLimit(...)`, leaving `newNonceCache` referenced only from `_test.go` call sites, so golangci-lint's `unused` linter now flags it as dead code. Please either delete `newNonceCache` and call `newNonceCacheWithByteLimit` directly from the three test call sites, or otherwise remove the now-dead wrapper.

internal/signer/api/requestsign_test.go
```diff
@@ -194,30 +206,63 @@ func TestNonceCache_Expiry(t *testing.T) {
...
+		t.Fatalf("retained key bytes = %d, exceeds budget %d", c.bytes, 2*nonceCacheKeyBytes)
+	}
+	if err := c.checkAndStore("caller", largeNonce+"3"); err != errNonceCacheFull {
+		t.Fatalf("third large nonce should hit the byte budget, got %v", err)
+	}
+	for key := range c.entries {
+		if len(key) != nonceCacheKeyBytes {
+			t.Fatalf("cache retained key width %d, want %d", len(key), nonceCacheKeyBytes)
+		}
+	}
```
arepala-uml, line 261: Non-blocking (agrees with CodeRabbit's nitpick on this head): `key` is `[nonceCacheKeyBytes]byte`, an array type, so `len(key)` is a compile-time constant equal to `nonceCacheKeyBytes` — this condition can never actually fail, so the loop can't catch a key-width regression. Consider asserting `len(c.entries)` and/or `c.bytes` instead, as CodeRabbit's suggested diff does.

internal/signer/api/requestsign.go
```diff
@@ -180,53 +191,83 @@ func (a *RequestSigningAuthenticator) readBody(r *http.Request) ([]byte, error)
...
-	entries map[string]int64 // identifier -> expiry (unix nanos)
-	ttl     time.Duration
-	max     int
-	now     func() time.Time
+	mu       sync.Mutex
+	entries  map[[nonceCacheKeyBytes]byte]int64 // digest -> expiry (unix nanos)
+	ttl      time.Duration
+	max      int
+	bytes    int64
+	maxBytes int64
```
arepala-uml, line 203: Non-blocking observation: since every retained key is now fixed-width (`nonceCacheKeyBytes`), `c.bytes` is always exactly `len(c.entries) * nonceCacheKeyBytes` — the byte budget is mathematically redundant with the entry-count cap whenever `maxBytes == max * nonceCacheKeyBytes` (the production default via `defaultNonceCacheMaxBytes`). It only does independent work when constructed with a smaller custom `maxBytes`, as the tests do. Not incorrect — the digest itself is what fixes the unbounded-retention bug — just flagging that the extra accounting is inert in the shipped default path, in case that wasn't the intent.

arepala-uml (changes requested):
Verified the core fix locally: on origin/main, an attacker-controlled X-Nonce header (bounded only by the HTTP header-size limit, effectively up to ~1MB) was retained verbatim as the map key for up to defaultNonceCacheMax (65536) live entries — a worst-case tens-of-GB retention. This PR hashes (caller, nonce) into a fixed 32-byte SHA-256 digest with length-delimited fields (preventing caller/nonce boundary collision), capping worst-case retention at ~2MB. TTL (2*skew) and the entry-count cap are unchanged, so the replay-protection window/guarantee is not weakened — only the per-entry footprint changes. Tests (TestNonceCache_BoundsRetainedNonceBytes, TestRequestSigning_LargeNonceUsesBoundedReplayStorage, TestNonceCacheKey_DelimitsCallerAndNonce) exercise this correctly, and `go test -race ./internal/signer/...` passes.

One blocker: the head commit fails the required `lint` check (reproduced locally, see inline comment).

internal/signer/api/requestsign_test.go
wolf31o2: Fixed in f05a619 — you and CodeRabbit are both right that it guarded nothing.

Replaced with the invariant that can actually drift: retained bytes stay exactly the entry count times the key width, however large the nonces that produced those entries were. That still fails if the accounting stops matching what is stored, which is what the loop was reaching for.

internal/signer/api/requestsign.go
wolf31o2: Correct, and intended. With the digest in place the byte budget is redundant in the shipped default, where maxBytes == max * nonceCacheKeyBytes — it does independent work only under a smaller custom limit, as the tests use.

Keeping it deliberately: it is what makes the bound explicit at the point where retention happens rather than implicit in the key type, so a later change to a variable-width key (or to the default ratio) is caught by the accounting instead of silently reintroducing unbounded retention. The test now asserts exactly that relationship, so if it ever stops holding it fails rather than passing quietly.

arepala-uml (approved):

PR #858, opened 2026-09-07
test(extension): cover provider method contracts

wolf31o2:
## Summary

- Cover forwarding and results for all CIP-30 provider methods.
- Cover all CIP-95 methods and negotiated API behavior.
- Verify method names and parameters at the provider boundary.

## Validation

- Extension tests, type-check, lint, and build pass.
- Restrictive-CSP Chromium E2E was attempted but failed before provider registration; no browser-runtime result is claimed.

This adds contract coverage without changing production behavior.

---
## Summary by cubic
Adds contract tests for all CIP-30 and CIP-95 provider methods, verifying method names, parameters, and forwarded results at the extension boundary. `getCollateral` now returns `null` per the CIP-30 contract, and the E2E guide and grant-store tests now document and enforce exact HTTP(S) origins (rejecting `file://`, paths, and non-HTTP URLs).

<sup>Written for commit ce9a6038855648a1a93a99d41dc6eacae1b52117. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/bursa/pull/858?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/ui/extension/src/injected.ts
+++ b/ui/extension/src/injected.ts
@@ -38,7 +38,7 @@ interface CIP30API {
   getNetworkId(): Promise<number>;
   getUtxos(amount?: string, paginate?: Paginate): Promise<string[] | null>;
   getBalance(): Promise<string>;
-  getCollateral(params?: { amount: string }): Promise<string[]>;
+  getCollateral(params?: { amount: string }): Promise<string[] | null>;
   getUsedAddresses(paginate?: Paginate): Promise<string[]>;
   getUnusedAddresses(): Promise<string[]>;
   getChangeAddress(): Promise<string>;
--- a/ui/extension/tests/injected.test.ts
+++ b/ui/extension/tests/injected.test.ts
@@ -16,8 +16,20 @@ interface TestExtension {
 interface TestCIP30API {
   getExtensions(): Promise<TestExtension[]>;
   getNetworkId(): Promise<number>;
+  getUtxos(amount?: string, paginate?: { page: number; limit: number }): Promise<string[] | null>;
+  getBalance(): Promise<string>;
+  getCollateral(params?: { amount: string }): Promise<string[]>;
+  getUsedAddresses(paginate?: { page: number; limit: number }): Promise<string[]>;
+  getUnusedAddresses(): Promise<string[]>;
+  getChangeAddress(): Promise<string>;
+  getRewardAddresses(): Promise<string[]>;
+  signTx(tx: string, partialSign?: boolean): Promise<string>;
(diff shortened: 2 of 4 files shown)


ui/extension/tests/injected.test.ts
```diff
@@ -99,6 +189,99 @@ describe('injected provider', () => {
     expect(api.cip95).toBeUndefined();
   });
 
+  it('forwards and resolves every CIP-30 and enabled CIP-95 method', async () => {
```
arepala-uml, line 192: Confirmed: this test's CIP-30 coverage (getNetworkId/getUtxos/getBalance/getUsedAddresses/getUnusedAddresses/getChangeAddress/getRewardAddresses/signTx/signData/submitTx, plus the three cip95 methods) is re-asserted almost verbatim by the two tests below at lines 500 and 604 — same methods, same params, same source/method assertions, just different literal result values. The one thing unique to this test is asserting `getCollateral` resolves to `null` (the case the two tests below don't cover, since their `getCollateral` case uses a non-null array). Non-blocking: agree with cubic's duplication finding. Suggest folding the null-result assertion for `getCollateral` into the `forwards every CIP-30 method with its contract parameters` test (line 500) and dropping the rest of this test, rather than maintaining three near-identical contract lists.

ui/extension/src/injected.ts
```diff
@@ -38,7 +38,7 @@ interface CIP30API {
   getNetworkId(): Promise<number>;
   getUtxos(amount?: string, paginate?: Paginate): Promise<string[] | null>;
   getBalance(): Promise<string>;
-  getCollateral(params?: { amount: string }): Promise<string[]>;
+  getCollateral(params?: { amount: string }): Promise<string[] | null>;
```
arepala-uml, line 41: Non-blocking follow-up (not introduced by this line, which is otherwise correct): the concrete implementation of this method at `buildCIP30API()` (line ~175-176) still declares `Promise<string[]>` and force-casts its result with `as Promise<string[]>`, so it never advertises the `null` that this interface now documents as part of the CIP-30 contract. It's harmless today — `api` is exposed to callers through this `CIP30API`-typed binding, so external consumers still see `string[] | null` — but the two declarations of the same contract can drift; consider updating the implementation's own return type and dropping the narrowing cast.

arepala-uml (approved):

PR #871, opened 2026-09-13
fix(ui): show each piece of bootstrap work separately

wolf31o2:
## The problem

The bootstrap screen shows one bar, but the node is never doing one thing. It runs work in parallel at two levels and reports all of it through a single progress field, so that one bar alternates between unrelated percentages over unrelated totals.

**Several downloads at once.** A fresh-sync run sampled at 5s resolution caught three of them inside the one `bootstrap` phase, live in `/status`:

```
digest list          3.4 MB   100%
ancillary ledger    244 MB   86.7%   5.2 MB/s
immutable archives  13.8 GB    9.3%  31.7 MB/s
```

`bootstrap.total_bytes` — the single field the screen read — flips between `3525541`, `255611612` and `14780304000` from poll to poll. dingo fetches these in parallel (*"Steps 4+5: Download immutable archives and the ancillary archive in parallel"*, `mithril/bootstrap_v2.go`) and labels each report with the artifact it describes, *"so concurrent downloads can be distinguished by callers consuming one callback"* — but drops that label when it flattens into `mithril.SyncProgress`.

**Two phases at once.** Later, the ledger import and the immutable copy run as two goroutines in one errgroup (`mithril/sync.go`). Measured across consecutive polls:

```
22:01  immutable_copy  75.1%
22:04  ledger_import   52.3%
22:07  immutable_copy  79.0%
22:11  ledger_import   78.1%
```

Each series advances on its own; only the display jumped. The phase checklist, which inferred each phase's state from its position in a fixed list, drew `immutable_copy` as *pending* while it sat at 79%.

**A 0% blip at every phase end.** dingo marks a phase's end with a report carrying no measurements, and `toBootstrapProgress` dropped the flag that says so — writing 0% over a phase that had just reached 99.8%.

## The change

Keep the progress of each piece of work separately, identified by phase **and** download size. The size is fixed for a download and different for each, so it is the identity still available once dingo's artifact label is gone. (Two downloads of exactly equal size would share a row; nothing worse.)

- `BootstrapProgress.Done` carries the phase-end edge instead of discarding it.
- `Status.BootstrapPhases` holds each piece of work's latest report, in first-seen order, merged copy-on-write so a snapshot already handed out is never mutated under its reader. The end edge completes every download under that phase at 100%, keeping what each measured.
- A phase's bare "started" report is a placeholder: it stands while it is the only thing known, and the phase's first measured report takes its place rather than leaving an empty row beside the real ones.
- The screen renders one row per piece of work — own bar, own percent, own bytes, rate and ETA. Work that has finished stays visible as finished, which is what keeps the next phase starting at 0.1% from reading as lost progress.
- Where a phase holds more than one download, each row is named by size (`Download snapshot · 13.8 GB`); two rows reading `Download snapshot` would look like a rendering fault rather than like two downloads. A lone download keeps the plain phase name.
- The checklist takes each phase's state from what the node reported, and keeps a phase active while any of its downloads is still running.
- The sync banner names what is still working, dropping anything finished — it is one line, and three entries wrap to three lines at phone width.

## Screenshots

**The parallel downloads**, rendered from the values `/status` served during the run above:

live parallel downloads

Phone width:

phone width

**Before, one poll** — a single bar reading 79.0%:

before a

**Before, the next poll** — the same moment now reads 52.3%, and "Copy chain history" is drawn as not-yet-started although it is 79% done:

before b

**After**, at the concurrent-phase stage — both phases with their own numbers, and the finished download still shown as finished:

after

## Follow-on for the node

dingo already knows which download each report belongs to and labels it for exactly this purpose; only the flattening into `mithril.SyncProgress` loses it. Carrying `Artifact` through would let this name the downloads ("chain archives", "ledger state") instead of inferring identity from their size. Worth a separate issue there.

## Note on the first commit

The first commit on this branch added a "Step N of 7" position next to the percent, assuming the phases run in sequence and the percent merely resets at each handoff. The measured run disproved that — the phases overlap, so a step counter reads 3, then 2, then 3 as they take turns reporting. The later commits remove it and fix the actual cause.

## Tests

Eleven Go tests and thirteen web tests, each written failing first: concurrent downloads kept apart by size, concurrent phases kept apart, the phase opener replaced by the first measured report, the phase end completing every download under it rather than blanking one, `Status()` copying the slice rather than handing out the live one, per-row bars and their accessible names, size in the label only where a phase holds more than one download, the banner dropping finished work, the checklist's state coming from reports rather than position, and the single-report fallback.

`go build ./...`, the full `ui` Go suite and `nilaway` on the changed package are clean; `881 passed (69 files)` on the web suite, plus `npm run build`, `type-check` and `lint`.

### DIFF
--- a/ui/web/src/bootstrapPhases.ts
+++ b/ui/web/src/bootstrapPhases.ts
@@ -25,3 +25,22 @@ export function bootstrapPhaseLabel(phase: string): string {
     phase.replace(/_/g, " ")
   );
 }
+
+// Where a phase sits in the pipeline, or null for a phase we do not know.
+//
+// The per-phase percent restarts at 0 on every handoff. Measured on preview
+// (dingo v0.70.5): the snapshot download read 99.8% one moment and the next
+// phase read 0.1% five seconds later, with the byte readout replaced by a block
+// count. Showing the step position next to that percent is what keeps the reset
+// legible as forward motion rather than a restart.
+//
+// dingo may skip a phase — that run went straight from "bootstrap" to
+// "immutable_copy" — so the step can jump. This reports a position in the
+// pipeline rather than a count of phases seen, so a skip reads as progress.
+export function bootstrapStepPosition(
+  phase: string,
+): { step: number; total: number } | null {
+  const idx = BOOTSTRAP_PHASES.findIndex((p) => p.key === phase);
+  if (idx < 0) return null;
+  return { step: idx + 1, total: BOOTSTRAP_PHASES.length };
+}
--- a/ui/web/src/components/SyncBanner.tsx
+++ b/ui/web/src/components/SyncBanner.tsx
@@ -1,6 +1,6 @@
 import type { Tone } from "./StatusPill";
 import type { Status, NodeState } from "../api/types";
-import { bootstrapPhaseLabel } from "../bootstrapPhases";
+import { bootstrapPhaseLabel, bootstrapStepPosition } from "../bootstrapPhases";
 
 interface SyncBannerProps {
   status: Status;
@@ -27,7 +27,16 @@ export function SyncBanner({ status }: SyncBannerProps) {
 
   let detail = "";
   if (status.state === "bootstrapping" && status.bootstrap) {
+    // The percent belongs to the current phase alone, and every handoff
+    // restarts it near zero. Naming the step the percent is measuring is what
+    // keeps that reset legible as the next step starting rather than as lost
+    // progress; this strip is on screen for the whole bootstrap, so it is where
+    // the collapse is seen most. A phase we cannot place gets no step.
+    const position = bootstrapStepPosition(status.bootstrap.phase);
     detail = `${bootstrapPhaseLabel(status.bootstrap.phase)} ${status.bootstrap.percent.toFixed(1)}%`;
+    if (position) {
+      detail += ` · Step ${position.step} of ${position.total}`;
+    }
   } else if (status.state === "ready") {
     detail = `tip ${status.tip} · ${status.caughtUp ? "caught up" : "catching up"}`;
   } else if (status.error) {
--- a/ui/web/src/components/components.test.tsx
+++ b/ui/web/src/components/components.test.tsx
@@ -414,3 +414,74 @@ test("MultiSigProgress shows threshold-met indicator once signedCount reaches th
   expect(screen.getByText(/2 of 2/)).toBeInTheDocument();
   expect(screen.getByText(/threshold met/i)).toBeInTheDocument();
 });
+
+// The percent in the banner measures the CURRENT phase, and every handoff
+// restarts it near zero (measured on preview: 99.8% -> 0.1% in five seconds).
+// Without the step position the banner reads as progress being thrown away —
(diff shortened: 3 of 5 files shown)


ui/internal/supervisor/supervisor.go
```diff
@@ -407,7 +407,86 @@ func (s *Supervisor) onProgressForRun(runID uint64, bp BootstrapProgress) {
 	if !s.activeRunLocked(runID) || s.status.State != StateBootstrapping {
 		return
 	}
-	s.status.Bootstrap = &bp
+	s.status.BootstrapPhases = mergeProgress(s.status.BootstrapPhases, bp)
+	// The headline is whatever just reported, taken from the merged entry so
+	// the empty end edge reads as "finished" rather than blanking work that had
+	// just measured its way to 99.8%.
+	for i := range s.status.BootstrapPhases {
+		if sameWork(s.status.BootstrapPhases[i], bp) {
```
arepala-uml: Stale headline after a multi-download phase ends. `mergeProgress` marks every retained entry of a phase `Done: true` on its end edge, matching by `Phase` alone (line 445). This loop instead matches by `sameWork` (`Phase` **and** `TotalBytes`), but the end edge carries `TotalBytes: 0` — so for any phase with nonzero-`TotalBytes` entries (concurrent downloads, the scenario this PR is centrally about), no entry matches and `s.status.Bootstrap` is never refreshed. It stays frozen at whichever single download last reported, with `Done: false`.

Reproduced directly: three downloads under phase `bootstrap` (`TotalBytes` 100/200/300) followed by its `Done: true` end edge leaves every entry in `BootstrapPhases` correctly `Done: true`, but `status.Bootstrap` unchanged at `{TotalBytes:300, Done:false}`.

`ui/web/src/components/SyncBanner.tsx`'s fallback (`shown = running.length > 0 ? running : [status.bootstrap]`) triggers exactly in this window — right after every phase entry finishes and before the next phase reports — so the banner shows a stale non-done percentage instead of a sensible transition state. `Syncing.tsx` is unaffected since it reads `bootstrap_phases` directly.

Suggest matching this loop by `Phase` alone (like the `mergeProgress` done-branch) rather than by `sameWork`.

arepala-uml (changes requested):
Confirmed one bug with a reproduction; everything else checked out (build/vet/lint/nilaway clean, 881 frontend tests passing, no bot findings).

ui/internal/supervisor/supervisor.go
wolf31o2: Confirmed and fixed in fb2f95f — thank you, the diagnosis was exact.

I reproduced it as you described: three downloads under `bootstrap` (100/200/300) plus the end edge left every entry `Done: true` but the headline frozen at `{TotalBytes:300, Done:false}`. That is now a regression test.

Rather than matching by `Phase` alone in that loop, I removed the second lookup entirely — `mergeProgress` returns the headline along with the merged list, so the matching happens once, where the entry is already in hand. A second lookup keyed differently from the merge is what made this possible in the first place.

For an end edge the headline is the phase's last entry: entries are in first-seen order, so for the download phase that is the main archive rather than the small files fetched alongside it.

arepala-uml (approved):

PR #898, opened 2026-09-19
feat(ui): add tray update checks

wolf31o2:
## Summary

- Add a `Check for updates` action to the desktop wallet tray menu.
- Check the latest stable Bursa GitHub release and notify the user when an update is available.
- Open only the validated Bursa release page when an update is found.

## Scope

This is a manual update check; it does not download or replace wallet binaries.

---
## Summary by cubic
Adds a manual update check to the desktop wallet tray. The tray previously had no update path; it now compares the embedded build version against the latest stable Bursa GitHub release and notifies when an update is available, without downloading or replacing wallet binaries.

- Development builds (no embedded version) report the latest release only and never announce an update.
- Update pages open only when the release URL and version tag pass validation, restricting URLs to the official Bursa release path.
- Adds a `version` package with build-time `Version` and `CommitHash` injected via linker flags.

<sup>Written for commit fb483898212bf378ba51dd3fefd154e015abe2f2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/bursa/pull/898?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added `bursa key bls` for generating Dijkstra-era BLS stake-pool keys and registration materials.
  - Added Cardano CLI-compatible signing and verification key files.
  - Added desktop tray support for checking wallet updates and opening available releases.
  - Added build version and commit information to the webview application.

- **Documentation**
  - Expanded key derivation guidance, including BLS key generation, signing envelopes, registration JSON, and key rotation details.

- **Bug Fixes**
  - Added validation for invalid BLS data and duplicate output paths.
  - Added safeguards for untrusted update release URLs.

PR #886, opened 2026-09-17
Polish responsive wallet UI and add design previews

JaeBrianLee:
Polishes the wallet interface across Portfolio, Send, Receive, Activity, Staking and Settings while preserving Bursa’s slate-and-gold palette and existing capability/signing gates. Adds responsive navigation, clearer token rows, upright transfer surfaces and more readable address/settings layouts.

Uses Bursa’s existing pouch logo and custom wordmark from `.github/assets` throughout the branded surfaces. Includes self-hosted Manrope, a standalone sample-data preview and 14 desktop/mobile screenshots. Receive preserves the current `usage_known` safeguard. No backend changes.

## Validation

- Production build passes; 905 tests across 70 files pass.
- ESLint: zero errors; one existing unused-disable warning in `connectorNotify.ts`.
- Browser checks: token search, address copy, QR expansion, transfer fields, native-asset entry, delegation setup, keyboard tabs and mobile navigation.
- No document overflow across the five task pages at 320px. Desktop and mobile screenshots refreshed after rebasing onto current main.
- Screenshots use synthetic data. Transactions and settings writes are blocked in the preview; live-node flows were not exercised.

## Screenshots

Mobile screenshots show the actual 390 × 844 viewport. Staking includes a second view after scrolling; its fixed navigation stays at the viewport bottom.

<details>
<summary>Portfolio — desktop and mobile</summary>

Portfolio desktop

Portfolio mobile

</details>

<details>
<summary>Send — desktop and mobile</summary>

Send desktop

Send mobile

</details>

<details>
<summary>Receive — desktop and mobile</summary>

Receive desktop

Receive mobile

</details>

<details>
<summary>Staking — desktop and mobile</summary>

Staking desktop

Staking mobile

Staking mobile after scrolling

</details>

<details>
<summary>Activity — desktop and mobile</summary>

Activity desktop

Activity mobile

</details>

<details>
<summary>Settings — desktop and mobile</summary>

Settings desktop

Settings mobile

</details>

[Screenshot index and preview instructions](https://github.com/JaeBrianLee/bursa/blob/faa2bc08704b2ff9eaf758f73952aa1f25605682/docs/design/README.md)

## Summary by CodeRabbit

- **New Features**
  - Redesigned the wallet interface with a slate-and-gold visual theme, branded typography, icons, and updated navigation.
  - Added responsive layouts, including mobile bottom navigation and improved layouts for portfolio, sending, receiving, staking, activity, and settings.
[...]

PR #900, opened 2026-09-20
fix(mobile): exclude the iOS wallet tree from iCloud backup

wolf31o2:
`Library/Application Support` is included in iCloud and iTunes backup by
default, and `isExcludedFromBackup` appeared nowhere under `mobile/`. The iOS
wallet tree therefore left the device in a backup the app never opted out of.

## What lives under `dataDir`

`ui/internal/boot/boot.go` creates, under the directory the shell passes:
`db/` (the embedded Dingo chain database, the multi-gigabyte part),
`node.socket`, `settings.json`, `contacts.json`, `multisig.json`,
`keystore.json` (legacy), and `vault.json`.

`vault.json` is the wallet key material. Nothing is in the Keychain. Per
`ui/internal/vault/vault.go` it is encrypted at rest: a per-vault VEK sealed
under the vault unlock phrase, the wallet index sealed under the VEK, and each
seed sealed separately under its own spending phrase, all scrypt + AES-256-GCM.

**The call: exclude the whole tree, key material included.** Excluding it means
a device restore does not bring the wallet back, and that is the intended
trade-off — recovery for this wallet is the recovery phrase, which the create
flow already makes the user write down and re-enter. The alternative, leaving
`vault.json` in iCloud, puts an offline-attackable scrypt blob in a backup to
buy a recovery path that duplicates one the user already has. Excluding only
`db/` would also require the shell to pre-create a directory the Go layer owns.

`Library/Caches` is the wrong home for the same reason: the system may purge it
at any time, and `vault.json` cannot be regenerated by re-syncing. Application
Support plus `isExcludedFromBackup` is the combination that matches this data.

## Both trees, not just the destination

`dataDir` is excluded immediately after `createDirectory` and before the legacy
migration copies anything in, so the destination is never eligible for a backup
that begins mid-copy.

`Documents` is backed up too, and the legacy tree outlives the copy: it is
still there during the copy window, after a cleanup a launch deferred, and
after a name `removeMigratedLegacyEntries` could not remove on a later launch.
So it carries the same exclusion, set in the same place and guarded only on the
directory existing — not only on the migration-failure fallback path. Nothing
else uses that directory: the app declares neither `UIFileSharingEnabled` nor
`LSSupportsOpeningDocumentsInPlace`, so excluding it takes nothing of the
user's out of backup.

Exclusion is not a first-creation-only step: the flag is a per-URL resource
value, so an install predating this change, or a directory restored from an
older backup, has the directory present without it. Re-setting an
already-excluded URL is a no-op, so the call is idempotent and unconditional.

## When the exclusion itself fails

`setResourceValues` can fail — a disk too full to write the extended attribute
while the multi-gigabyte chain database is being copied is the likely case —
and the result is a wallet tree that is still backup-eligible, which is the
condition the call exists to remove.

`excludeFromBackup` returns the failure description rather than only logging
it, every call site routes it through `noteBackupExclusionFailure`, and
`viewDidAppear` presents it as its own one-time alert, ahead of the migration
alert when both fired. Only the first failure of a launch is kept: the
condition that fails one `setResourceValues` fails the rest.

**Not fatal, and not retried within the launch.** The wallet's data is intact
in every case this can occur, the flag is re-set on the next launch anyway, and
refusing to launch would deny access to a working wallet over a condition the
app cannot clear — the user is the one who frees the disk. Retrying in-launch
would re-attempt the same call against the same full disk. What it must not do
is stay in the system log, which is not somewhere a phone user looks.

## Migration failure

The `catch` returned `documentsDir ?? dataDir`, silently running the node from
the directory the app is migrating away from.

Kept as a fallback rather than made a hard launch failure. The legacy tree is
emptied only after the copy is verified, so on failure it is still the
authoritative copy while the destination holds nothing but the rolled-back
partial copy: booting against the destination would show an empty wallet, and
the next launch would then discard whatever the user created in it. Refusing to
launch instead would deny access to a wallet whose data is intact, over a
condition the next launch retries.

What changed is that it is no longer silent, and no longer unprotected. The
fallback directory carries the exclusion, and the failure is surfaced to the
user once on first appearance rather than only in the system log.

## Android

`mobile/android/app/src/main/AndroidManifest.xml` sets
`android:allowBackup="false"` (since blinklabs-io/bursa#561), which opts out of
cloud backup and restore. That attribute does not govern Android 12+
device-to-device transfer.

So the manifest now also sets
`android:dataExtractionRules="@xml/data_extraction_rules"`, and
`mobile/android/app/src/main/res/xml/data_extraction_rules.xml` excludes every
domain — `root`, `file`, `database`, `sharedpref`, `external` — from
`<device-transfer>`. The domains are separate extraction trees, so excluding
`root` alone does not cover the others. The same exclusions are repeated under
`<cloud-backup>`, so flipping `allowBackup` can never silently re-enrol the
wallet in a cloud backup. The file follows the existing
`res/xml/network_security_config.xml` pattern.

The platform documents that on devices from some manufacturers app files cannot
be kept out of D2D migration at all, so the transfer exclusion is best effort;
the cloud opt-out is the one the platform guarantees. `mobile/README.md` says
so. The rules file is ignored on Android 11 and lower, which have no D2D
transfer path and where `allowBackup="false"` is the whole opt-out.

## Verification

`mobile/verify-contracts.sh` checks the exclusion at both call sites, the
legacy-tree exclusion inside the range between `createDirectory` and the copy
loop, the failure return from `excludeFromBackup`, the failure reaching the
alert through `viewDidAppear` and `presentStartupWarnings`, that every
`Self.excludeFromBackup` call is wrapped in `noteBackupExclusionFailure`, the
fallback reporting itself, the Android `allowBackup` opt-out, the manifest
`dataExtractionRules` attribute, and an exclude-only `<device-transfer>`
section.

Each new check was shown to fail on its own. With the previous head's
`WalletViewController.swift` restored, the script exits 1 at the legacy-tree
check. Dropping the exclusion result at the `dataDir` call site, at the
fallback call site, making `excludeFromBackup` return `nil` on failure, cutting
`viewDidAppear`'s read of the failure, or removing the `present(alert,
animated: true)` call each exit 1. Removing the manifest attribute, deleting
the rules file, and adding an `<include>` inside `<device-transfer>` each exit
1 as well.

Merged against `origin/main` at `1177a81` (after blinklabs-io/bursa#899): the
merged tree differs from `origin/main` only in this change's five files and
from this branch only in that pull request's files, so nothing was reverted by
the merge. `verify-contracts.sh`, `go build ./...`, `go vet ./...` and `go test
./...` all pass on the merged tree. No other open bursa pull request
(blinklabs-io/bursa#886, #898, #901, #902, #903, #525) touches any path under
`mobile/`, so the file-set intersection is empty.

**This change is unbuilt and unrun.** No Xcode, simulator, device, or Android
SDK was available, and no CI job compiles either file: `verify mobile
contracts` is a ripgrep source-contract check, and the `mobile` workflow is
`workflow_dispatch` with the `android` job gated on `BURSA_ANDROID_ENABLED`. A
reviewer on a Mac must verify:

1. It compiles: `cd mobile/ios && xcodegen generate && xcodebuild -scheme Bursa
   -sdk iphonesimulator build`, against the iOS 15.0 deployment target in
   `project.yml`. The new `StartupWarning` nested struct, the
   `presentStartupWarnings` recursion, the `String?` return from
   `excludeFromBackup`, and the `guard let detail,` shorthand binding are the
   parts that have never seen a compiler.
2. `NSURLIsExcludedFromBackupKey` reads true on `Library/Application
   Support/Bursa` **and** on `Documents` after a launch that has a legacy tree,
   and still reads true on the second launch of an app installed before this
   change.
3. With a migration failure simulated, the alert presents once and does not
   re-present on subsequent foregrounding; with both a failed exclusion and a
   failed migration, the two alerts present in sequence rather than the second
   being dropped for presenting on an already-presenting controller.
4. Android: `./gradlew assembleDebug` accepts `data_extraction_rules.xml` —
   aapt2 validates the domain values — and Android lint raises nothing new.

Adjacent prior work: blinklabs-io/bursa#852 moved the iOS tree to Application
Support and added the marker-guarded migration this builds on. No open issue
covers mobile backup.

---
## Summary by cubic
Previously, iCloud and iTunes backed up the iOS wallet tree because `Library/Application Support` is included by default. The app now marks its data directory `isExcludedFromBackup` on every launch, so the chain database and encrypted wallet stay on the device; a device restore won't bring the wallet back, and recovery is via the recovery phrase.

- The legacy `Documents` tree is excluded before the migration copy begins, not only on the failure fallback, and a failed exclusion is surfaced in a one-time alert instead of only the system log.
- Android already set `allowBackup="false"`; a new `data_extraction_rules.xml` excludes every domain from `<device-transfer>` so Android 12+ device-to-device transfer cannot carry the wallet either.
- `verify-contracts.sh` now checks the iOS call sites, the alert surfacing, and the Android manifest and rules. This change is unbuilt and unrun, so a Mac reviewer should confirm it compiles and the flag takes effect after launch.

<sup>Written for commit 2d9d57bb05ce72a8e2a2cc8e9a9607011bfb538e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/blinklabs-io/bursa/pull/900?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved wallet data migration handling by safely falling back to the legacy wallet directory when migration fails.
  - Added a one-time alert explaining when migration fallback occurs.
  - Wallet data is excluded from iCloud and Android backups, helping keep the wallet on the device and requiring recovery via the recovery phrase after device restoration.

- **Documentation**
  - Documented wallet backup behavior and verification coverage for iOS and Android.


mobile/ios/Bursa/WalletViewController.swift
```diff
@@ -91,6 +114,9 @@ class WalletViewController: UIViewController, WKNavigationDelegate {
             try fileManager.createDirectory(
                 at: dataDir, withIntermediateDirectories: true
             )
+            // Before anything is written or copied in, so the tree is never
+            // eligible for a backup that starts mid-migration.
+            Self.excludeFromBackup(dataDir)
```
cryptodj413: documentsDir is never passed to excludeFromBackup: the legacy tree stays backup-eligible during the copy window before removal, when the post-marker removeItem loop fails and returns early, and when removeMigratedLegacyEntries cannot remove a recorded name on a later launch; add Self.excludeFromBackup(documentsDir) here, before the copy begins, mirroring the dataDir call.

mobile/ios/Bursa/WalletViewController.swift
```diff
@@ -159,9 +185,37 @@ class WalletViewController: UIViewController, WKNavigationDelegate {
...
+    //
+    // Application Support + this flag, rather than Library/Caches: the system
+    // may purge Caches at any time, and vault.json is the encrypted wallet
+    // seeds, which cannot be regenerated by re-syncing.
+    private static func excludeFromBackup(_ url: URL) {
+        var url = url
+        var values = URLResourceValues()
+        values.isExcludedFromBackup = true
+        do {
+            try url.setResourceValues(values)
```
cryptodj413, line 288: setResourceValues can fail here too, for example on a full disk while copying the multi-gigabyte chain database, and the only signal is a log line; route this failure through migrationFailureDetail as well so an unexcluded wallet tree is reported to the user the same way a failed migration already is.

mobile/README.md
```diff
@@ -172,6 +172,14 @@ device can be checked with:
...
+`allowBackup` opt-out, and the iOS Application Support data directory with its
+backup exclusion. It does not replace Android emulator, physical-device, iOS
+simulator, or physical-device validation.
+
+The iOS wallet tree lives in `Library/Application Support/Bursa`, which iCloud
+and iTunes back up by default. The app marks that directory
+`isExcludedFromBackup` on every launch, so the chain database and the encrypted
+vault stay on the device. A device restore therefore does not carry the wallet
+over; recovery is from the recovery phrase. Android sets
+`android:allowBackup="false"` for the same reason.
```
cryptodj413: allowBackup="false" stops iCloud and iTunes backup but does not cover Android 12+ device-to-device transfer, which is governed separately by android:dataExtractionRules; this app defines none, so add that caveat to this sentence.

cryptodj413 (changes requested):
Verified the documentsDir backup-exclusion gap and the excludeFromBackup failure path directly against the diff; both hold at head 238cb115. One inline comment is a required change, two are non-blocking recommendations on lines this PR already touches.

mobile/ios/Bursa/WalletViewController.swift
wolf31o2: Fixed in 2d9d57b: Self.excludeFromBackup(documentsDir) now runs immediately after the dataDir call, before the marker branch and before the copy, guarded only on the directory existing, so the legacy tree is excluded during the copy window, after a cleanup a launch deferred, and after a name removeMigratedLegacyEntries could not remove. verify-contracts.sh asserts it inside the createDirectory-to-copy-loop range, and restoring 238cb11's WalletViewController.swift exits 1 on exactly that check.

mobile/ios/Bursa/WalletViewController.swift
wolf31o2: Fixed in 2d9d57b: excludeFromBackup returns the failure description, every call site routes it through noteBackupExclusionFailure, and viewDidAppear presents it as its own one-time alert ahead of the migration one. Not fatal and not retried in-launch: the data is intact, the flag is re-set every launch, and refusing to launch would deny access to a working wallet over a full disk the app cannot clear. Contract checks cover the return, the alert path, and that every exclusion call reports; mutating any one exits 1.

mobile/README.md
wolf31o2: Fixed rather than deferred in 2d9d57b: mobile/android/app/src/main/res/xml/data_extraction_rules.xml excludes every domain from <device-transfer>, wired through android:dataExtractionRules, with the same exclusions under <cloud-backup> so flipping allowBackup cannot silently re-enrol the wallet. The README carries the caveat as well, because the platform documents that some manufacturers' devices cannot keep app files out of D2D migration at all. Unbuilt here: no Android SDK, and the android job is workflow_dispatch-gated.

cryptodj413 (approved):

